PR #67, opened 2014-05-29
Retrieve shared formulas

Y--:
Implement 18.3.1.40


matthewmueller:
What ended up happening with this PR?

jasconius:
Hi, bumping this. I've made a private fork of XLSX JS and have created a true blue formula parser and resolver... and I'm just now running into this little issue... the shared formula thing is a big issue and I am wondering if anyone has any documentation about the rules that govern shared formulas and how I can make inferences from the data in the XML and generate the appropriate formulas. I'm pretty much ready to build this, I just need more data to pursue it.

Doing the corporate dance to try and get my work on parsing/resolving into your repo, not easy, but I think you guys would be quite happy with what I've created.

SheetJSDev:
This was addressed in 0.8.4.  XLSX XLSB XLS and SpreadsheetML now parse shared and array formulae.

In XLSX, the raw shared formula is stored in the upper left cell along with a shared id `si` attribute.  Cells referencing the formula include that `si` attribute.  The parser calculates a delta and applies it to the relative references in the formula.

The current iteration uses a very simple regex which only fail on some special cases that can be refined with a more precise regular expression.  @jasconius I originally thought a proper parser was needed but in retrospect a fine tuned regular expression can work (Excel itself prohibits defined names that potentially collide with a valid cell address).

PR #116, opened 2014-09-18
Update xlsx.js

OlafurTorfi:
JSZip written with capitals in jszip.js and lower case in xlsx.js...


lrdevries:
Je problem isn't that JSZip is or isn't in capitils, the problem is that in bower.json jszip isn't a dependency of js-xlsx. Therefore if you handle your dependencies automatic, the order of the files isn't correct, because it loads js-xlsx before jszip.

reviewher:
@lrdevries the bower entry point has been changed to a script that embeds the jszip dependency directly.

Closing because I think the underlying issue was resolved, please raise a new issue if it is still a problem

PR #161, opened 2015-01-15
Workbook.DefinedName(s) support has been added.

alitskevich:
From this fork https://github.com/alitskevich/js-xlsx

### DIFF
--- a/bits/77_wbxml.js
+++ b/bits/77_wbxml.js
@@ -16,7 +16,7 @@ function parse_wb_xml(data, opts) {
 		return r;
 	}
 
-	// discover defined names
+	// Discover defined names
 	wb.DefinedNames = [];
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Generally looks good, not in front of computer at the moment so I can't test just yet.  Few comments:

1) can you lift the regex definitions out of the function?  if you look just above your change, wbnsregex is lifted out of parse_wb_xml so that it's processed at load time

2) can you squash down to one commit?

3) does `make lint` pass?

alitskevich:
Thank you for fast feedback (it is my very first PR on github),
1) not sure, because call re.exec()  for //g modifying re.lastIndex.
2-3) ok, i'll check with lint and prepare a new PR tomorrow
4) seems it also needs to add utf8read(attrValue)

Issue #177, opened 2015-02-20
Incorrect handling of table:number-columns-repeated in ODS


ziacik:
Hi. I think there may be incorrect handling of number-columns-repeated attribute in ODS files.

If I have a table where two adjacent cells have the same value, the second will not be present in Sheet object.

For example, having an ODS with `A1` cell set to "A" and `B1` cell set to "A" too,

`console.log(require('xlsx').readFile('./Test.ods').Sheets.Sheet1);`

outputs this:

`{ A1: { t: 's', v: 'A', w: 'A' }, '!ref': 'A1' }`

So B1 is missing there.

Version I tried this on is 0.7.12.

SheetJSDev:
@ziacik do you have a sample file so I can take a look?

ziacik:
@SheetJSDev Sample file is at this link http://www.koliksoftware.net/Stuff/Test.ods

ziacik:
Merged cells are not handled well either as you can see from http://www.koliksoftware.net/Stuff/Test2.ods

JavaScriptDude:
What is holding this back? This bug makes this toolset unusable for parsing ODS files. The patch works for me.

Dineshagrawal85:
i'm facing same problem is there any solution for this?

JavaScriptDude:
I had to fork this locally and apply ziacik's patch. Ugly but only way to get around this without involvement of owners.

PR #179, opened 2015-02-23
Fix repeated columns handling in ODS.

ziacik:
Hi. Please review the changes I believe fix issue #177 
Sorry for not including some unit tests.


SheetJSDev:
@ziacik The cell object `q` needs to be cloned.  Try reading the file and changing one of the repeat values -- you will see that the other cells also change value (the file does not say that the cells are linked, only that they happen to share the same value)

ziacik:
Hi. Added simple cloning of the cell object.

JavaScriptDude:
I have confirmed that this patch works for my tests

Issue #165, opened 2015-01-26
Needs to add a dependency in bower.json ?


vincentcialdella:
Hi,

I need to specifiy the url of jszip.js file in dist folder else I have this error: undefined is not a function
It concerns line 5093 of xslx.js file: jszip is undefined.

If I add a dependecy <script src="bower_components/js-xlsx/dist/jszip.js"></script>, this is working.

As this file seems to be a mandatory dependency, should not you update bower.json file like this : 

{
  "name": "js-xlsx",
  "homepage": "https://github.com/SheetJS/js-xlsx",
  "main": ["dist/jszip.js","dist/xlsx.js"],
  "version": "0.7.11",
  "ignore": [
    "bin",
    "bits",
    "misc",
    "*_/._"
  ],
  "keywords": [
    "excel",
    "xlsx",
    "xlsm",
    "xlsb",
    "ods",
    "js-xlsx"
  ]
}

Thank you for your answer.

PR #180, opened 2015-02-23
Add dependency in bower to packaged jszip

owencraig:
Fixing issue #165. This will enable devs to just bower install the package without also needing to install jszip (in the correct order)


forcewill:
+1

jeanp:
+1

Issue #81, opened 2014-07-08
How to access Row Object and set properties


neversaid:
Hi All,

thanks for creating such a great project.
I tried to figure out how its possible to set some Row Properties when creating a XLSX Spreadsheet - i need to set the property "hidden" on the Row to hide it.
How is the best practice to do this?

Thanks a lot,
Bernhard

PR #188, opened 2015-03-04
handle !rows array

paulish:
Add an ability to set !rows array where each row can contain hidden
property and height property in pixels (hpx) or points (hpt).
Add an utility function to convert pixels to points.

### DIFF
--- a/bits/45_styutils.js
+++ b/bits/45_styutils.js
@@ -62,3 +62,5 @@ function find_mdw(collw, coll) {
 		if(MDW === MAX_MDW) MDW = DEF_MDW;
 	}
 }
+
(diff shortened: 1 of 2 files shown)


paulish:
The patch only writes !rows to xlsx. The reading and testing parts are missing. The patch fixes https://github.com/SheetJS/js-xlsx/issues/81

jkurei:
What is the state on this issue?

I'm trying to write an xlsx with higher rows, and this code isn't working: 

```
sheet['!rows'] = Array(10).fill({hpx: 100}) // (hpt doesnt work either)
```

A very similar code, with `wpx`, is working to make columns wider, but no luck with rows.

PR #194, opened 2015-03-29
disambiguate duplicate header names

patricklam:
If multiple columns have the same header, rename subsequent columns to HEADER_1, HEADER_2, etc.

### DIFF
--- a/bits/90_utils.js
+++ b/bits/90_utils.js
@@ -84,6 +84,7 @@ function sheet_to_json(sheet, opts){
 	var cols = new Array(r.e.c-r.s.c+1);
 	var out = new Array(r.e.r-r.s.r-offset+1);
 	var outi = 0;
+	var hdrSet = {};
 	for(C = r.s.c; C <= r.e.c; ++C) {
 		cols[C] = encode_col(C);
 		val = sheet[cols[C] + rr];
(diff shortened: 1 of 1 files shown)


SheetJSDev:
@patricklam so if I'm reading the code correctly (haven't tested it), if the table is

Foo | Bar | Bar | Bar | Bar
------------ | ------------- | ------------ | ------------- | ------------
1 | 2 | 3 | 4 | 5

then the row object will be

```json
{
    "Foo": 1,
    "Bar": 2,
    "Bar_2": 3,
    "Bar_3": 4,
    "Bar_4": 5
}
```

Do we want to rename the first Bar to Bar_1 ?

@notatestuser @nathanathan thoughts?

patricklam:
The first ambiguous one will be "Bar_1" not "Bar_2" right now. That's kind
of like what happens when you download files and you have a duplicate,
which then gets called File (1) or whatever. I'm indifferent as to what it
is exactly, just have to pick something. It would be another 5 minutes of
implementation to rename the first Bar.

YueZhaoDreams:
Is this implemented in the current version? I have a concern that when the headers are [Bar, Bar, Bar_1] will the fix work fine?

laurence-hudson-tessella:
+1

r1z1a:
Is there way to disable this?

SheetJSDev:
If you are running into this issue @r1z1a that means your data file has multiple column headers with the same value, like [the example from an earlier post in the thread](https://github.com/SheetJS/js-xlsx/pull/194#issuecomment-87346537):

| Foo | Bar | Bar | Bar | Bar |
| --- | --- | --- | --- | --- |
| 1 | 2 | 3 | 4 | 5 |

The old behavior used the last column, so you would have seen `{Foo:1,Bar:5}`.  The current behavior disambiguates so you would see `{Foo:1,Bar:2,Bar_1:3,Bar_2:4,Bar_3:5}`.  If you really want to do it, you can generate custom headers in an array and pass it to sheet_to_json, as shown in the README: `XLSX.utils.sheet_to_json(ws, {header:['Foo', undefined, undefined, undefined, 'Bar']})`

PR #210, opened 2015-04-21
Force sheet_to_json to include empty cells

instantaphex:
You may want to leave this as an option, but this was necessary for my use case.


instantaphex:
Sorry about that.  I only wanted to make a pull request for 7b7e954

PR #255, opened 2015-06-15
Fixing error with missing .Props file

martinheidegger:
`zip.file` uses `wb.Props` but if `.Props` is undefined it will throw an error. Seems like the safeguard from a few lines below should be before `zip.file`


SheetJSDev:
Thanks @martinheidegger a bunch of related potentially-undefined tests were added

PR #256, opened 2015-06-15
Issue 147 fix

jtadmor:
Proposed solution to Issue 147 (where a column in which the first row is an empty cell is dropped, even if later rows in the column are non-empty).

Allow user to pass in parameter parseEmpty: true (or something truthy).

In sheet_to_json, this will create a val of '' where a cell is empty, rather than going straight to continue. To guard against empty rows being added, isEmpty is still not set to true unless the val is truthy.


SheetJSDev:
Closed in latest commit, the options field to set the default value is `defval`

PR #271, opened 2015-07-08
Fix SST parsing when SST XML has a tag prefix

morstaine:
XLSX documents written by DocumentFormat.OpenXml.dll (at least sometimes) contain a tag prefix "x" in the sharedStrings.xml file. This is a fix to the regular expressions that parse this file to handle the prefix if it is present.


SheetJSDev:
@morstaine thanks for looking into this! There were quite a few places where the code assumed the XML tags would not be namespaced, not just in the shared string table, and generally Excel 2007-2013 do not write namespaces when they are not needed.

PR #282, opened 2015-08-06
added compression option to the write file

bchr02:
By adding a compression option to writeFile this will now allow us to override the default behavior of JSZip which is to STORE (no compression). 

Example usage:
XLSX.writeFile(wb, 'out.xlsx', {compression: 'DEFLATE'});


bchr02:
I will issue a new pull request. Updating the files within the bits directory.

PR #316, opened 2015-11-04
Fix/issue148/unsupportedvaluetype

ziacik:


SheetJSDev:
We ended up adding something similar to your solution in 0.8.2.  We had a different way of duplicating cells, so I'll pull that in and mark the commit.

PR #340, opened 2015-12-30
Modularity fix & sheet_to_csv option

araymer:
Fixed the issue in xlsx.js which was causing bundlers to be unable to use and added an "ignoreBlankRows" option that sheet_to_csv() can take so that it skips rows which are entirely blank.


reviewher:
We've improved the bundler usage.  There are other pull requests covering the blank rows, and I think a better approach is to just keep track of whether the row is blank rather than regexing at the end

PR #351, opened 2016-01-13
substr() doesn't work properly with IE8

Aymkdn:
Use slice() instead of substr() to make it work with IE8


SheetJSDev:
@Aymkdn all bad cases where substr was being used with a negative index were replaced with slice.  This also resolved some IE6 issues with reading XLSB files!

Issue #333, opened 2015-12-17
special format Excel to Json with error


xushuheng0623:
There is something wrong. special format Excel to Json 
[test.xlsx](https://github.com/SheetJS/js-xlsx/files/64848/test.xlsx)

unrecognized type stub

abhishek1234321:
did anybody get this working?

mateuszkrzeszowiec:
Don't think so, seem to be facing this problem :(

Issue #50, opened 2014-03-06
Blank cells are ignored.


ghost:
Blank cells in rows ['value1', ' ', 'value3'] are parsed into an object that does not recognize the blank cell: { A1: {}, C1: {} }

Is this the intended behavior? 

If it is, it would be nice to have an option that would include blank cells in the output from .readFile().Sheets.

If it is not, is there a way to read blank cells as well?

Thanks

hmnshmshr:
Solution 1 .Condition "if(h===undefined)continue;"  in  "xlsx.core.min.js" comment it out.

 or do it properly...  

Solution 2 . By passing Condition extra param while running this XLSX.utils.sheet_to_json(wb.Sheets[name] , {blankCell : false}). add a condition on line no. 19150   "if(defval === undefined && blankCell) continue;" in file xlsx.js etc..

ashishdutta007:
I tried with XLSX.utils.sheet_to_json(wb.Sheets[name] , {blankCell : false}) but i am still getting blank values for the empty cell

hmnshmshr:
Solution 1 .Condition "if(h===undefined)continue;" in "xlsx.core.min.js"
comment it out.
 add a condition on line no. 19150 "if(defval === undefined && blankCell)
continue;" in file xlsx.js etc..

Himanshu Mishra
Aiims Jodhpur

UI5Training:
Hi,

Did any of you able to find its solution?
I am facing the same problem.

YesicaMacias14:
Hi, did any one know some way to insert that data into a table?. The issue is I already catch the info from my excel file but I want to see it on my web side and insert them on a table in my database.

GaneshSinghPapola:
anyone find  the solution?

backstab319:
> anyone find the solution?

You can use the defval: "" option while reading a doc, and it will work.

`
xlsx.utils.sheet_to_json(
      workbook.Sheets[workbook.SheetNames[0]], {
        defval: ""
      });
`
Like this, hope it helps.

PR #354, opened 2016-01-18
Fix export of styled blank cells to JSON

Martin-Pitt:
Here are two fixes that support the ability to export the cellStyles of blank cells into JSON.

One fix involves allowing cellStubs to exist in JSON. (90_utils.js:117)
Second fix involves formating cellStyles correctly on stub cells. (66_wscommon.js:38)

This is for reading files with `cellStyles` & `sheetStubs` both set and then trying to convert a worksheet with `XLSX.utils.sheet_to_json`.

I'm not an employee of Microsoft or signed any of their NDAs, etc.

This may be related to: https://github.com/SheetJS/js-xlsx/issues/333, https://github.com/SheetJS/js-xlsx/issues/50
I tried searching for other relevant issues but was unable to find them. I would have originally put the cellStyles as an issue but decided to go ahead and just simply fix it myself and do a pull request.

I haven't been able to run the tests, complains about some mocha thing. The changes are super simple though so I am sure that someone else could do some verification tests that have their environment already working and ready.


reviewher:
@Martin-Pitt the whole stub processing was revamped and I think the underlying issues were resolved.

PR #363, opened 2016-02-04
Addition for write protected sheets

Mior:
I have added write protection for xml sheets using password hash from http://www.openoffice.org/sc/excelfileformat.pdf format.
It was tested and its working well. It is nothing but addition of one tag <sheetProtection> to sheet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,12 @@ Special worksheet keys (accessible as `worksheet[key]`, each starting with `!`):
   will write all cells in the merge range if they exist, so be sure that only
   the first cell (upper-left) in the range is set.
 
+- `ws['sheetProtection']`: object of write sheet protection properties. One of
+  [autoFilter, deleteColumns, deleteRows, formatCells, formatColumns, formatRows,
+  insertColumns, insertHyperlinks, insertRows, objects, pivotTables, scenarios,
+  selectLockedCells, selectUnlockedCells, sheet, sort, password]. Where most
+  most important is password which creates password hash to write protect sheet.
+
 ### Workbook Object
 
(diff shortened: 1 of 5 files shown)


lsholubka:
Hello, any updates on this? That would be a great feature to have.

SheetJSDev:
The relevant sections of the official specs include

- ECMA-376 Section 18.3.1.85 for XLSX
- MS-XLSB BrtSheetProtection record
- MS-XLS Protect and XF records
- MS-OFFCRYPTO password hashes

Keys attached to the worksheet object should be prefixed with an exclamation point (like `!ref` or `!cols`).

SheetJSDev:
Started digging into this, looks straightforward but we're going to have to make some changes (I'll rewrite the commit):

- According to MS-XLSB, the sheet protection info should show up after the data table but before the merge cells.  The write line was put in the right position

- The worksheet key should be `!sheetProtection`

- There are [two helper functions](https://github.com/SheetJS/js-xlsx/blob/master/bits/22_xmlutils.js#L150-L151) for generating the XML: `wxt_helper` will generate the attribute string and `writextag` will generate the xml tag given a tag name, body, and properties object; and `wxt_helper`.  In this case you could just use writextag on the sheetProtection object.

- According to [MS-XLSB](https://msdn.microsoft.com/en-us/library/dd944996.aspx), the actual password verifier algorithm is based on [MS-OFFCRYPTO] section 2.3.7.1, Binary Document Password Verifier Derivation Method 1.  Incidentally [we have that algorithm!](https://github.com/SheetJS/js-xlsx/blob/master/bits/44_offcrypto.js#L60).

PR #374, opened 2016-02-29
Fix incorrectly used options variable in readSync

kenspirit:
Passed in opts is incorrectly used instead of the internal variable o
derived.


reviewher:
That entire block was split into parts based on input type, and from a cursory glance it looks like the options are now used correctly.

Issue #383, opened 2016-03-28
ODS: Cells with 'number-columns-repeated' are not duplicated


mdamt:
As the subject says, the cells are not duplicated.

When printed as CSV:

```
a,v,c,d,e,,
1.01,aa,NULL,1,1
1.01,aa,NULL,2 <--- WRONG
```

it should be

```
a,v,c,d,e,,
1.01,aa,NULL,1,1
1.01,aa,NULL,NULL, 2 <--- the NULL is duplicated
```

Test file is available here: https://cloudup.com/cKV3Jpvfh34

PR #384, opened 2016-03-28
Fix: #383 - Duplicate cells with number-columns-repeated attribute

mdamt:
This patch duplicates cells when they have 'number-repeated-columns' attribute specified.


SheetJSDev:
We addressed this in 0.8.2.  In this case the object needs to be duplicated rather than merely referencing the same object.

PR #414, opened 2016-05-17
add icv{Top|Bottom|Left|Right}

ronnywang:
add icvTop, icvBottom, icvLeft, icvRight for border position


SheetJSDev:
Based on the relevant sections 2.5.20 CellXF and 2.5.246 StyleXF, I believe the icvLeft calculation is incorrect -- it should be `(b >> 16) & 0x7F`.

We're going to fill out that entire section in the next release

Issue #159, opened 2015-01-11
sheet_to_json doesn't output anything for a blank cell


marcelosomers:
It seems like using sheet_to_json doesn't handle blank cells very well. I want to keep them in my script (I'm parsing an xlsx file and outputting a csv), but a blank cell outputs an array like:

```
[ 'foo',
  'bar',
  ,      // the blank value
  'another value' ]
```
It'd be nice if you got a blank string, since this library seems to recognize the blank value. Would this be possible?

SheetJSDev:
The code as it stands now won't generate an empty string.  The simplest way is to just add a post-processing loop to your code:

```
for(var i = 0; i != json.length; ++i) for(var j = 0; j != json[i].length; ++j) if(typeof json[i][j] === 'undefined') json[i][j] = "";
```

As for what it "should" be, why not the number 0?  an empty string would be weird in a case like:

```
[
  [1, 2, 3],
  [4, 5, 6],
  [7, ? ,9]
]
```

I'd be amenable to adding an option (maybe something called "default" in the options object?), and the changes are relatively simple: https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L5293

```
			val = sheet[cols[C] + rr];
			if(val === undefined || val.t === undefined) continue;
```

In the bottom half of sheet_to_json, we walk in row-major order across the worksheet.  The first line attempts to find if the sheet has a cell at the prescribed address.  If the cell doesn't exist or if its type is not set, then we skip the cell.  Instead of just continuing, that second line should set the cell to the empty string:

```
			if(val === undefined || val.t === undefined) { row[hdr[C]] = ""; continue; }
```

marcelosomers:
Wow, really appreciate the lightning quick response! Your for loop snippet worked perfectly - I'm probably not experienced enough to comment on the intricacies of what it "should" be, but thank you!

tmthyjames:
Changing line 11549 from 
`if(val === undefined || val.t === undefined) continue;`
to
`if(val === undefined || val.t === undefined) { row[hdr[C]] = ""; continue; }`
fixed the issue for me.

IgorShch:
@SheetJSDev Thank you for the suggestion. It fixed the issue for me too. 

It would be really nice to turn this into an option. 

Thanks a lot.

bhaskarsontakke:
Great work! You saved my day. Thanks alot :)

prasanthvel:
Me too having same problem.
   


[...]

PR #420, opened 2016-05-29
Added "blankValue" option to sheet_to_json function

ljelewis:
This change adds an option to the second argument that is used when a blank cell is encountered in the sheet. The default behaviour, which remains in place when the new option is not present, is that blank cells are entirely omitted from the returned JSON object.

To use the option, include a "blankValue" key to the 2nd argument object and give it a value (other than false or undefined). The value is used as the value in the returned JSON object for any blank cells in the sheet.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -11504,7 +11504,7 @@ function format_cell(cell, v) {
 }
 
 function sheet_to_json(sheet, opts){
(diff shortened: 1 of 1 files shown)


ljelewis:
This change relates to issue #159

amenadiel:
I would totally use this feature.

SheetJSDev:
The option is `defval`, just added in commit 9ae3a64 -- it will be available in the 0.9.6 release

PR #422, opened 2016-06-08
wbrels[i] can be null

mihailik:
wbrels

Excels can have various structures populated in unusual ways. I've noticed this error in the field.

Putting a small check should protect against it.


reviewher:
@mihailik do you have a sample file that you can share?

reviewher:
This was fixed in 0.8.3 along with a bunch of related null checks

PR #437, opened 2016-07-11
Adding sheetFormatPr for better compat with macos preview

seriousben:


reviewher:
@SheetJSDev  We need to decide where to store the default row height and column width in the worksheet ...

SheetJSDev:
We will eventually overhaul the whole thing, but we will commit something that writes default height/width in XLSX if you pass `sheetFormat`

PR #471, opened 2016-09-28
commented out the line that set the length of buffer

husseiny-m:
https://github.com/DataGarage/node-xls-json/issues/14


reviewher:
This was fixed in 0.8.1 when the codepage dependency was updated

PR #486, opened 2016-10-21
Update README.md

nanachimi:
reference to library xlso added for parsing workbook object in JS object


nanachimi:
@SheetJSDev I don't know your check are failing. I just add a comment in a README.

PR #503, opened 2016-11-18
Setting 'state' attribute in 'sheet' tag so that a sheet can be hidden or made visible.

Digity:
Fixed issue where state attribute of sheet was being ignored and not included in generated xml.  Setting the state attribute allows users to hide or make a sheet visible.


ManjunathSVBB:
Still not able to hide the sheet

PR #534, opened 2017-01-22
Fix #533 (Buffer deprecation notice)

### LINKED ISSUE #533: buffer interface deprecation
line 1443 in xlsx.js

	var utf8readc = function utf8readc(data) { return Buffer(data, 'binary').toString('utf8'); };

needs to become 

	var utf8readc = function utf8readc(data) { return (new Buffer(data, 'binary')).toString('utf8'); };

for newer versions of nodejs

bbatliner:
The exact message from Node 7.0.0: ```DeprecationWarning: Using Buffer without `new` will soon stop working. Use `new Buffer()`, or preferably `Buffer.from()`, `Buffer.allocUnsafe()` or `Buffer.alloc()` instead.```

Change proposed in #533, PR here. Thanks.


bbatliner:
Deprecation notice removed (for now).

PR #541, opened 2017-01-25
support custom format dates such as "[$-409]mmm\-yy;@"

rjmcguire:


SheetJSDev:
This change makes little sense because the cells would be processed anyway if you scroll down (`else p.w = ...`).  The purpose of the `w`/`v` distinction is to save both the raw value and the display text.  If you can share a file that isn't properly handled, we can take a look, but as it stands this isn't a good fix.

PR #557, opened 2017-02-13
fixing an issue where _xd_ is getting escaped

fcrick:
\_x000D\_ style replacements always have exactly 4 hexadecimal characters.

If in Excel, you type \_xd\_ in a cell, js-xlsx will give this string special treatment, even though Excel doesn't consider it to be significant.If you enter \_x000D\_ in a cell in Excel, the leading underscore will be escaped by Excel as without that it would be treated as an escape.

### DIFF
--- a/bits/22_xmlutils.js
+++ b/bits/22_xmlutils.js
@@ -32,7 +32,8 @@ var rencstr = "&<>'\"".split("");
 
 // TODO: CP remap (need to read file version to determine OS)
 var unescapexml/*:StringConv*/ = (function() {
-	var encregex = /&[a-z]*;/g, coderegex = /_x([\da-fA-F]+)_/g;
+	/* 22.4.2.4 bstr (Basic String) */
+	var encregex = /&[a-z]*;/g, coderegex = /_x([\da-fA-F]{4})_/g;
 	return function unescapexml(text/*:string*/)/*:string*/ {
 		var s = text + '';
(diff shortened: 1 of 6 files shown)


SheetJSDev:
Good catch! This part is actually explained in ECMA-376 22.4.2.4 bstr:

> Unicode characters that cannot be directly represented in XML as defined by the XML 1.0 specification, shall be escaped using the Unicode numerical character representation escape character format _xHHHH_, where H represents a hexadecimal character in the character's value. [Example: The Unicode character 8 is not permitted in an XML 1.0 document, so it shall be escaped as _x0008_. end example] To store the literal form of an escape sequence, the initial underscore shall itself be escaped (i.e. stored as _x005F_). [Example: The string literal _x0008_ would be stored as _x005F_x0008_. end example]

Can you do the following:

- restart from the current master

- make the change to `bits/22_xmlutils.js` and `odsbits/22_xmlutils.js`

- run `make` and confirm that the same changes appear in `xlsx.js` and `ods.js`

- just above that line, in the function body, add a comment ` /* 22.4.2.4 bstr (Basic String) */`

We'll need to delve further into both the write and read sides to make sure everything is consistent, but for now that will suffice.

fcrick:
this what you had in mind?

SheetJSDev:
@fcrick looks good, check your email

Issue #439, opened 2016-07-15
Error while parsing XLSX with one of the file's name starting with uppercase letter inside


chikh:
I have a XLSX file containing `xl/SharedStrings.xml` inside. Trying to run:

``` js
require('xlsx').readFile(path)
```

leads to the error: `Cannot find file xl/sharedStrings.xml in zip`. The problem is that my SharedStrings starts with uppercase letter but the parser tries to get sharedStrings which begins with lowercase letter.

Please, consider my PR to solve the issue: https://github.com/SheetJS/js-xlsx/issues/439
Another PR: https://github.com/protobi/js-xlsx/pull/46

reviewher:
@chikh do you have a sample file? For this error to occur, the `[Content_Types].xml` file must be pointing to `xl/sharedStrings.xml` with a lowercase `s`

SheetJSDev:
This is in fact a bug.  Even though writers should be case-sensitive, readers should be case-insensitive according to ECMA-376 Part 2 Section 10.1.2 "Mapping Content Types".

The lay of the land is as follows:

| reader      | Handling files where [Content_Types].xml case differs from ZIP |
|-------------|----------------------------------------------------------------|
| Excel 2016  | Case insensitive for all files                                 |
| LibreOffice | Case insensitive for all files                                 |
| Numbers     | Case sensitive for some parts, not for sharedstrings.xml       |
| G Sheets    | Case sensitive for some parts, not for sharedstrings.xml       |
| Numbers IOS | Case sensitive for all files                                   |

Instead of duplicating the object and doing a check, it would be more efficient to check key by key

PR #567, opened 2017-02-22
Improved safegetzipfile()

duzun:
I have an xlsx file that can't be read by this lib because there is a `xl/SharedStrings.xml` file (notice upper case "S").

I've improved `safegetzipfile()` function to overcome this issue.

### DIFF
--- a/bits/21_ziputils.js
+++ b/bits/21_ziputils.js
@@ -24,7 +24,12 @@ function getdata(data) { return (data && data.name.slice(-4) === ".bin") ? getda
 function safegetzipfile(zip, file/*:string*/) {
 	var f = file; if(zip.files[f]) return zip.files[f];
 	f = file.toLowerCase(); if(zip.files[f]) return zip.files[f];
-	f = f.replace(/\//g,'\\'); if(zip.files[f]) return zip.files[f];
+	var g = f.replace(/\//g,'\\'); if(zip.files[g]) return zip.files[g];
+	f = keys(zip.files).filter(function (n) {
+		var m = n.toLowerCase();
+		return zip.files[n] && (m == f);
+	});
+	if ( f.length == 1 ) return zip.files[f[0]];
 	return null;
 }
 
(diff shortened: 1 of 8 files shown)


SheetJSDev:
@duzun thanks for the heads up!  It's a strange error because I can manually produce files with a capital S that work in excel and when fed back to js-xlsx.

If you don't mind doing a bit of sleuthing, can you look at the `[Content_Types].xml` file and find the part name corresponding to the content type `application/vnd.openxmlformats-officedocument.spreadsheetml.sharedStrings+xml`?  We look at the content types file to find the name of the shared strings file.  If that file was written correctly then you should see a line like:

```
<Override PartName="/xl/SharedStrings.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.sharedStrings+xml"/>
```

If the case is correct then there's some issue with the content types parsing.  If you see the part name has a different case, like `PartName="/xl/sharedStrings.xml"`, then we probably should make the getfile functions case insensitive.

reviewher:
Related: https://github.com/SheetJS/js-xlsx/issues/439

duzun:
Hi @SheetJSDev 

I've checked that file and found that `"application/vnd.openxmlformats-officedocument.spreadsheetml.sharedStrings+xml"` in `[Content_Types].xml` is `"/xl/sharedStrings.xml"`, but the file is named `"/xl/SharedStrings.xml"` in the archive.

I suppose this particular file is generated by some software other then office.

B.T.W. I've tested the new release `0.8.6` and the issue is gone! 

Thanks!

SheetJSDev:
@duzun either way, as I explained [in another comment](https://github.com/SheetJS/js-xlsx/issues/439#issuecomment-281903752), readers are supposed to do case-insensitive name comparisons.

And as for the PR, it's preferable to do a direct scan of the keys in this case.  I think there may be a few more places where `filter` is used and they should ultimately be reworked with a non-filter variant

PR #572, opened 2017-02-27
Fix make dist and browserify issue require

Liryna:


SheetJSDev:
Hello @Liryna !

The breaks in the require statements are used to trick browserify into not including those files.  If `require` is not available (like in the browser) none of those dependencies are pulled in.  When you are using browserify do you find it pulls in an FS shim or other shims?

Liryna:
Hi @SheetJSDev ,

It is not really browserify in my case, it is meteor that do his magic trick to detect which JS are used by scanning the require to send to the client side. With the break, it seems like meteor is unable to see it and the jszip is for example not sent and therefore missing at runtime.😢

SheetJSDev:
@Liryna the relevant changes were applied to un-break the require statements.  Please check if it works.  Keep the PR open though, since you also made some changes to the makefile I also want to revisit the build infrastructure

Liryna:
Hi @SheetJSDev ,

Thank you for the changes ! Yes it worked for me on meteor 👍

SheetJSDev:
We also folded the ODS logic into xlsx.js and reworked the cpexcel build to omit the unnecessary require, so I think we settled the require issues.

GNU sed doesn't accept a space between the `-i` and the specified extension.  The BSD sed appears to accept it, even though the manpage states the expected form is `-i extension`, so we will fix that as part of https://github.com/SheetJS/js-xlsx/issues/589

PR #581, opened 2017-03-03
Support of cellDates

Aymkdn:
Currently the `cellDates` option doesn't work and the Date cells are not correctly recognized.

I propose a fix that returns `.d` that contains the JavaScript Date.

I also propose to provide `fmt` and `fmtid` to the object so the user will be able to deal with any customized formats. It will give more flexibility.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -7850,8 +7850,124 @@ function get_cell_style(styles, cell, opts) {
 	return len;
 }
 
+/**
+ * This function permits to know if a "fmtid" is related to a date
+ * In other terms, it permits to know if a cell contains a date
+ * @param  {Number}  fmtid The fmtid of the cell
+ * @param  {String}  fmt   The format provided from fmtid using SSF._table[]
+ * @return {Boolean} TRUE if it's a Date, or FALSE if it's not
+ */
+function isFmtDate(fmtid, fmt) {
+  /* here is a list of fmtid from
+      - http://stackoverflow.com/questions/4730152/what-indicates-an-office-open-xml-cell-contains-a-date-time-value
+      - http://stackoverflow.com/questions/19034805/how-to-distinguish-inline-numbers-from-ole-automation-date-numbers-in-openxml-sp/19582685#19582685
+
+    (*) dates
+    0 = 'General';
+    1 = '0';
+    2 = '0.00';
+    3 = '#,##0';
+    4 = '#,##0.00';
+
+    9 = '0%';
+    10 = '0.00%';
+    11 = '0.00E+00';
+    12 = '# ?/?';
+    13 = '# ??/??';
+    14* = 'mm-dd-yy';
+    15* = 'd-mmm-yy';
+    16* = 'd-mmm';
+    17* = 'mmm-yy';
+    18* = 'h:mm AM/PM';
+    19* = 'h:mm:ss AM/PM';
+    20* = 'h:mm';
+    21* = 'h:mm:ss';
+    22* = 'm/d/yy h:mm';
+
+    27* = '[$-404]e/m/d';
+    28*
+    29*
+    30* = 'm/d/yy';
+    31*
+    32*
+    33*
+    34*
+    35*
+    36* = '[$-404]e/m/d';
+    37 = '#,##0 ;(#,##0)';
+    38 = '#,##0 ;[Red](#,##0)';
+    39 = '#,##0.00;(#,##0.00)';
+    40 = '#,##0.00;[Red](#,##0.00)';
+
+    44 = '_("$"* #,##0.00_);_("$"* \(#,##0.00\);_("$"* "-"??_);_(@_)';
+    45* = 'mm:ss';
+    46* = '[h]:mm:ss';
+    47* = 'mmss.0';
+    48 = '##0.0E+0';
+    49 = '@';
+    50* = '[$-404]e/m/d';
+    51*
+    52*
+    53*
+    54*
+    55* = 'yyyy/mm/dd';
+    56*
+    57* = '[$-404]e/m/d';
+    58*
+    59 = 't0';
+    60 = 't0.00';
+    61 = 't#,##0';
+    62 = 't#,##0.00';
+    67 = 't0%';
+    68 = 't0.00%';
+    69 = 't# ?/?';
+    70 = 't# ??/??';
+
+    164+ => customized format
+
+    Example of possible date formats:
+    '[$-F800]dddd\,\ mmmm\ dd\,\ yyyy'
+    'd/m;@'
+    'd/m/yy;@'
+    'dd/mm/yy;@'
+    '[$-40C]d\-mmm;@'
+    '[$-40C]d\-mmm\-yy;@'
+    '[$-40C]dd\-mmm\-yy;@'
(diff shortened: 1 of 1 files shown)


SheetJSDev:
What should `cellDates` do in general?

The original intention of the `cellDates` option was to normalize files for some ECMA-376 logic.  Excel generally writes files with dates stored as numbers (type `n`) with the date code, but the spec technically allows for storing ISO date strings (type `d`).  The default behavior is to convert the date cells to numeric, but setting `cellDates:true` preserves the native format.  On the write side, the default behavior is to convert cell objects with `t:"d"` to numeric cells.  Setting `cellDates:true` will save as XLSX date cells.

Note to self: SpreadsheetML and ODS should also follow the option

Aymkdn:
The user should be able to know when the cell contains a date. Right now nothing permits to know that, or I didn't understand how to proceed?! 

With the changes I did the intent is to indicate to the user that the cell contains a date and to parse it to the `.d`. So I'm opened to suggestions if you have another way to find this information.  

Also it looks important for the user to get the `fmt` in order to correctly treat the format of the content on his/her hand, don't you think? 

Thanks!

SheetJSDev:
> Also it looks important for the user to get the fmt in order to correctly treat the format of the content on his/her hand, don't you think?

The `cellNF` [option](https://github.com/SheetJS/js-xlsx/#parsing-options) should expose the actual number format string to the `z` field of the cell objects.  For example with the test file https://github.com/sheetjs/test_files/blob/master/number_format.xlsm?raw=true cell B24 of sheet "Implied" is set to the format `#,##0.00;[Red](#,##0.00)`:

```
> require('xlsx').readFile('number_format.xlsm').Sheets.Implied.B24
{ t: 'n', v: 12345.6789, w: '12,345.68' }
> require('xlsx').readFile('number_format.xlsm', {cellNF:true}).Sheets.Implied.B24
{ t: 'n',
  v: 12345.6789,
  w: '12,345.68',
  z: '#,##0.00;[Red]\\(#,##0.00\\)' }
```

A quick comparison shows the same behavior for [XLS](https://github.com/sheetjs/test_files/blob/master/number_format.xls?raw=true) and [XLSB](https://github.com/sheetjs/test_files/blob/master/number_format.xlsb?raw=true).

It does the same thing for the internal number formats, so you don't have to do a big switch on the format id -- you can just scan the string.

----

> The user should be able to know when the cell contains a date. Right now nothing permits to know that, or I didn't understand how to proceed?!

This should be something exposed by the `SSF` formatter, but you are right that the information is not available at the moment.

----

> With the changes I did the intent is to indicate to the user that the cell contains a date and to parse it to the .d. So I'm opened to suggestions if you have another way to find this information.

Disregard my last comment.  You want to proactively change number cells with a date format, and that's fine :)

---

Conclusion: 

1) try using the `cellNF:true` option and reading the format from the `.z` field of the cell.  If that works, we should not try to expose the internal number format ID.  

2) I will change the format parser so that it tells you if it thinks the format is a date.  Basically it would be a simplified form of the [tokenizer](https://github.com/SheetJS/js-xlsx/blob/master/bits/10_ssf.js#L527-L596)

Your commit is good, but the simple regex doesn't handle cases like the custom format `0 "months"` (since the "m" is in quotation marks it's a literal character).

Also, can you shoot an email to sheetjs@gmail.com -- the git log shows an anonymized address :/

Aymkdn:
Thanks for all these details. It helped me, and I'm now able to parse a date cell without modifying the original js-xlsx code by using `.z` with `cellNF:true` ! :-)

I know the way to parse the format is not perfect, but it should work in most cases for now.

This is for my [ExcelPlus library](http://aymkdn.github.io/ExcelPlus/) that I use as an interface of js-xslx to easily manipulate an Excel file.

reidblomquist:
@SheetJSDev I'm having issues getting a value in z on date objects coming in from an `xls` formatted file. They seem to just not be there... `ropts= { type: 'binary', cellNF: true }` is what I'm using (doing this all client side/in browser) - wondering if the file type I'm using or the binary flag is interfering with z value parsing or if you have any advice. Cheers!

SheetJSDev:
@reidblomquist if you aren't seeing cell formats, that's a bug in the library.  

I just checked against an [XLS file with a cell encoding a date](https://github.com/SheetJS/test_files/blob/master/xlsx-stream-d-date-cell.xls?raw=true) and it appears to populate the data for dates:

```js
> require('xlsx').readFile('xlsx-stream-d-date-cell.xls', {cellNF:true}).Sheets.Sheet1.B5
{ ixfe: 62,
  XF:
   { ifnt: 0,
     ifmt: 14,
     flags: 1,
     fStyle: 0,
     data: { patternType: null, icvFore: 64, icvBack: 65 } },
  v: 41684.35264774306,
  t: 'n',
  w: '2/14/14',
  z: 'm/d/yy' }
```

(note: the whole concept of the "date cell" type is specific to XLSX -- XLSB and XLS do not have a date cell type and always represent dates as numeric cells)

Can you share a file where `cellNF` is not adding formats to a cell? If you don't want to make it public, you can email us at dev@sheetjs.com

reidblomquist:
@SheetJSDev before I do that, I'm going to pull down the sheet that test uses and run it through my code. Would you also mind running a check using { type: 'binary' } in addition to cellNF true - wondering if that combo is what's causing. Thanks!

reidblomquist:
@SheetJSDev disregard! it is indeed an issue with formatting in the sheet my Ruby code is generating haha. Man sheet inception. Thanks again for being so prompt! Lmk if you want me to delete some of my comments on here to clean it up - and thanks for your hard work/awesome free software!

SheetJSDev:
@reidblomquist It's probably best to move this discussion to a new issue, as it is somewhat unrelated to the core issue of this PR and @Aymkdn probably doesn't want to see all of these unrelated notifications!

If you are generating a file that does the right thing in Excel (formats a cell properly) but not with js-xlsx, it's a bug in js-xlsx and I recommend raising a new issue with that file.

If your generated file isn't doing the right thing in Excel, then clearly your file generation process has an issue.

reidblomquist:
@SheetJSDev last comment here - but yes it does seem to be something with the file generation for the xls I'm reading from - I'll do some testing to confirm my findings but so far I'm thinking it has to do with the date not being formatted with traditional / delineators but with . notation. When I open my source xls file they're shown as Date types in excel, but if I re-select that format it reformats the cells to use / delineators and if I then run that "fixed" xls file through js-xlsx everything works as expected. If you'd like me to pull together some sample code and files and open up a new issue I'd be more than happy to!

SheetJSDev:
@reidblomquist yes that sounds like a different sort of bug in js-xlsx, so a new issue with a test file or some sample code would be appreciated :)  Once you raise that issue, I will go back and clean up these comments

SheetJSDev:
@Aymkdn we added an [`is_date` helper](https://github.com/SheetJS/ssf/blob/master/ssf.flow.js#L544) to SSF, factored out of the eval function.

Next version will have this fix and will ensure consistent `cellDates` handling

Aymkdn:
Awesome !!! Thank you 👍

SheetJSDev:
We expect to push the next release sometime this week

Issue #582, opened 2017-03-07
sheet_to_json empty cell be skip, suport to option is better?


jimjin:
excel:
Col1, Col2, Col3, Col4
Val1, Val2,          ,Val4

after call sheet_to_json():
`[{Col1: 'Val1', Col2: 'Val2', Col4:'Val4'}] # the key 'Col3' was lost.`

But, i need result is: [{Col1: 'Val1', Col2: 'Val2', Col3: null, Col4:'Val4'}]

If has option skipUndfendVale and defaultValue will better?
like: 
```
sheet_to_json({skipUndfendVale: false, defaultValue:null});
sheet_to_json({skipUndfendVale: false, defaultValue: ''});
sheet_to_json({skipUndfendVale: true});
```

jimjin:
And changed the soure code:
```
function sheet_to_json(sheet, opts){
  var val, row, range, header = 0, offset = 1, r, hdr = [], isempty, R, C, v;
  var o = opts != null ? opts : {};
  var skipUndefined = o.skipUndefined === undefined ? true : o.skipUndefined;
  var raw = o.raw;
  if(sheet == null || sheet["!ref"] == null) return [];
  range = o.range !== undefined ? o.range : sheet["!ref"];
  if(o.header === 1) header = 1;
  else if(o.header === "A") header = 2;
  else if(Array.isArray(o.header)) header = 3;
  switch(typeof range) {
    case 'string': r = safe_decode_range(range); break;
    case 'number': r = safe_decode_range(sheet["!ref"]); r.s.r = range; break;
    default: r = range;
  }
  if(header > 0) offset = 0;
  var rr = encode_row(r.s.r);
  var cols = new Array(r.e.c-r.s.c+1);
  var out = new Array(r.e.r-r.s.r-offset+1);
  var outi = 0;
  for(C = r.s.c; C <= r.e.c; ++C) {
    cols[C] = encode_col(C);
    val = sheet[cols[C] + rr];
    switch(header) {
      case 1: hdr[C] = C; break;
      case 2: hdr[C] = cols[C]; break;
      case 3: hdr[C] = o.header[C - r.s.c]; break;
      default:
        if(skipUndefined && val === undefined) continue;
        hdr[C] = format_cell(val);
    }
  }

  for (R = r.s.r + offset; R <= r.e.r; ++R) {
    rr = encode_row(R);
    isempty = true;
    if(header === 1) row = [];
    else {
      row = {};
[...]

Issue #147, opened 2014-11-19
excel read error when first row of a column is an empty cell.


kaustuv:
If excel data has a column whose value is an empty cell in the first row, then that column is not imported. This is possibly because utils.sheet_to_row_object_array() only has keys for non-empty cells.

This means that a simple example like this:

id col1 col2 col3
1 hello1        10
2 hello2 2 20
3 hello3 3 30

Only imports col1 and col3 correctly but not col2

zachhale:
I think I may be experiencing a similar issue which I haven't fully diagnosed yet. I have a spreadsheet where the first row is headers but the next two rows are completely empty and I am running into an issue where a value way down the sheet in row 513 has it's last column incorrectly parsed. The file in the data is 36071161600 but I'm getting 367120000000 after running the file through `xlsx`. After deleting the empty rows, the value comes in correctly. I'm working on diagnosing this into a proper test case still.

dominickp:
I'm also experiencing this issue. I can resolve it by adding data to the first couple of rows. Has anyone come up with a suitable workaround yet?

PR #583, opened 2017-03-07
resolved issue #582, to make method sheet_to_json options supported skipUndefined and defaultValue

jimjin:
Add options for sheet_to_json method:
```
sheet_to_json({skipUndefined: false, defaultValue: null});
sheet_to_json({skipUndefined: false, defaultValue: ''});
sheet_to_json({skipUndefined: true});
```


SheetJSDev:
I'm going to rework the pull as follows:

- the `skipUndefined` logic will be the default -- JSON writer generating the "undefined" key is a bug

- the `defaultValue` parameter will be renamed `default`.  If `default` is specified, missing data will be given that value.  If `default` is not specified the data will be skipped

Related issues: #582 h/t @jimjin , #147 h/t @jtadmor

SheetJSDev:
On second thought, it's bad to call it `default`, will stick to `defval`

PR #584, opened 2017-03-07
Update xlsx.js  

LieJiang:
write file whit bookSST true, it goes wrong,this can fix it.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -4474,7 +4474,7 @@ function write_sst_xml(sst, opts) {
 		if(s.r) sitag += s.r;
 		else {
 			sitag += "<t";
-			if(s.t.match(straywsregex)) sitag += ' xml:space="preserve"';
+			if((String)(s.t).match(straywsregex)) sitag += ' xml:space="preserve"';
 			sitag += ">" + escapexml(s.t) + "</t>";
 		}
 		sitag += "</si>";


SheetJSDev:
@LieJiang If the value is undefined it is most likely a stub cell.  The cell shouldn't be written in the first place.

Check your email

LieJiang:
If the conversion type is not enforced, it does not work with 'do not have the match function'

SheetJSDev:
@LieJiang I understand :) What I am saying is that the error is coming because of a bad cell.  For example:

```js
XLSX.write({SheetNames:["Sheet1"], Sheets:{Sheet1:{"!ref":"A1:A1", A1:{t:"s", v:3}}}},{bookType:'xlsx', bookSST:true, type:'buffer'});
XLSX.write({SheetNames:["Sheet1"], Sheets:{Sheet1:{"!ref":"A1:A1", A1:{t:"s", v:null}}}},{bookType:'xlsx', bookSST:true, type:'buffer'});
```

In the first case, the cell is a string cell with a numeric value. In this case, String will do the right thing, writing the text "3" to the cell.

In the second case, the cell is a string
cell with a null value.  If you use the `String` conversion, the output file will have the text `null` in those cells.  That is incorrect.  If the text is null it should not write the data in the first place.

This will require a few more changes to other parts of the code.  Keep the PR open

LieJiang:
We can add a filter before writing a file.

PR #594, opened 2017-03-14
Fix to vbaraw field is null when PartName attribute is /xl/vbaProject.bin

### LINKED ISSUE #348: vbaraw field is NULL even when bookVBA property is set to true
``` javascript
    <script>
        sforce.connection.query(
            "SELECT Id, Body FROM StaticResource WHERE Name = 'Sample_file'", 
            { onSuccess : handleUpdate,
              onFailure : handleFailure });
        function handleUpdate(result) {
            result = result.getArray('records');
            //console.log(result[0].Body);
            var workbook = XLSX.read(result[0].Body, {type:"base64",bookVBA:true});
            console.log(workbook.vbaraw);
        }
    </script>
```

Here, I am reading Excel file "Sample_file.xlsm" and trying to get the vba code present in the file by passing option parameter bookVBA as true but it returns null. 

<b>Note</b>: It is reading the contents of the excel perfectly but not the vba code.

Could somebody help me on this? I don't know what is wrong I am doing here. Thanks!

fossamagna:
fix #348

### DIFF
--- a/bits/85_parsezip.js
+++ b/bits/85_parsezip.js
@@ -135,7 +135,7 @@ function parse_zip(zip/*:ZIP*/, opts/*:?ParseOpts*/)/*:Workbook*/ {
 		out.files = zip.files;
 	}
 	if(opts.bookVBA) {
-		if(dir.vba.length > 0) out.vbaraw = getzipdata(zip,dir.vba[0],true);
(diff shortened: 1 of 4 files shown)


SheetJSDev:
@fossamagna good catch!  Since we're having some issues with travis (there's a 4MB limit to output), we had to change the test suite and rework the commit.

PR #629, opened 2017-04-13
Fix Cell Objects to always have html-encoded data for "h" key

xkr47:
Some cases left characters like `&` and `<` unencoded in the "h" key, making it unreliable for direct embedding in html.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -1553,6 +1553,10 @@ function escapexml(text, xml){
 	return s.replace(decregex, function(y) { return rencoding[y]; }).replace(charegex,function(s) { return "_x" + ("000"+s.charCodeAt(0).toString(16)).slice(-4) + "_";});
 }
 function escapexmltag(text){ return escapexml(text).replace(/ /g,"_x0020_"); }
+function escapehtml(text){
+	var s = text + '';
+	return s.replace(decregex, function(y) { return rencoding[y]; });
+}
 
 /* TODO: handle codepages */
 var xlml_fixstr = (function() {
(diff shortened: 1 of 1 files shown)


SheetJSDev:
What should happen with other entities?  Here's a fun test case we threw together for testing entities in plaintext formats: 

[CODES.xlsx](https://github.com/SheetJS/js-xlsx/files/920106/CODES.xlsx)

Excel's XML uses a special form `_xHHHH_` for encoding other entities, as demonstrated in B2 in Sheet1 from the attached workbook:

```xml
      <c r="B2" t="str">
        <f>CHAR(A2)</f>
        <v>_x0001_</v>
      </c>
```

SheetJSDev:
Squashed the two commits and added a few tests in https://github.com/SheetJS/js-xlsx/commit/616d2e534fc8736040fbb4fbdeb5ff97a515aad4

The open question is whether more HTML entities should be supported.  For example, take the unicode BLACK HEART SUIT ♥  There are two acceptable HTML encodings:

- `&hearts;` <html>&hearts;</html>
- `&#x2665;` <html>&#x2665;</html>

xkr47:
@SheetJSDev Thanks for adding tests!
If we are only talking about *encoding* html, here's my opinion:

I don't think it makes sense to encode more entities than is strictly necessary. Other tools outside xlsx can be used if more encoding is needed. I think entities were more popular when people didn't use UTF-8 but some other charset. Special Excel XML entities like _xHHHH_ should be decoded before encoding to HTML.

xkr47:
Thanks!

SheetJSDev:
@xkr47 there are some more corner cases to work out, but that's for another day.  0.9.11 also integrates an HTML writer that properly handles merge cells as rowspan/colspan.

Even though the PR shows up as "closed" rather than "merged", you are marked as the author of commit 616d2e5

xkr47:
Yeah I noticed, :+1:

PR #643, opened 2017-04-26
sheet_to_json with selected range & header: 1 has an array that starts at 0

plhmhck:
When converting with sheet_to_json and selecting a range and header: 1 the created array had some undefined entries at the start of the array.

e.g. B1:C2 would have an array where the first index with a value would be 1.

For a use case of mine I needed it to start always with 0, which I believe make more sense in general


SheetJSDev:
This is reasonable since it mirrors the behavior when the range skips the first row.  Check your email :)

PR #655, opened 2017-05-14
WIP: Add TypeScript definitions

wolfgang42:
This pulls in all of the definitions from [@types/xlsx](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/73c2d769b8bd95a6ff54ea2c43bf53fa7091b0e7/types/xlsx), plus the changes from DefinitelyTyped/DefinitelyTyped/pull/16500. I've extracted the relevant commits from the DT repo and added it as a new root so all the history is preserved.

I have renamed all of the interfaces so they no longer start with `I`, but I notice that the Flow types have similar but not quite the same names as the TypeScript ones. Would it make sense to use something like [joarwilk/flowgen](/joarwilk/flowgen) to autogenerate the definitions of one from the other? That way they would always be in sync.

The build system is somewhat baffling--I added the references to dtslint to `package.json` but I'm not certain that's where they belong, and I wasn't sure where the linting belonged in the testing process. Also, I don't have a great deal of experience with TypeScript. On the whole, the changes I've made probably ought to be reviewed before they are merged in.

CC @DMan577, @DxCx, @groetzi, @Knagis, @nickveys, @paulvanbrenk, @rinzeb, @RyanCavanaugh, @StasD, @themauveavenger, @tmburnell who have all been involved in the development of the DefinitelyTyped definitions.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,7 +1,5 @@
-// Type definitions for xlsx
 // Project: https://github.com/SheetJS/js-xlsx
 // Definitions by: themauveavenger <https://github.com/themauveavenger/>, Wolfgang Faust <https://github.com/wolfgang42>
-// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /** Attempts to read filename and parse */
 export function readFile(filename: string, opts?: IParsingOptions): IWorkBook;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,5 +1,6 @@
 // Project: https://github.com/SheetJS/js-xlsx
 // Definitions by: themauveavenger <https://github.com/themauveavenger/>, Wolfgang Faust <https://github.com/wolfgang42>
+// TypeScript Version: 2.2
 
 /** Attempts to read filename and parse */
 export function readFile(filename: string, opts?: IParsingOptions): IWorkBook;
@@ -377,12 +378,12 @@ export interface IWorkSheetCell {
     /**
      * Cell hyperlink object (.Target holds link, .tooltip is tooltip)
      */
-    l?: Object;
+    l?: object;
 
     /**
      * The style/theme of the cell (if applicable)
      */
-    s?: Object;
+    s?: object;
 }
 
 export interface ICell {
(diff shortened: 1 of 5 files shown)


SheetJSDev:
Wow thanks for doing this and keeping the original history of the type definitions!  The ideal solution would be the inverse of flowgen (pulling flow comments to generate TSD) but I didn't find any good guidance on that process.  Either way, it still makes sense to keep the definitions here.

From what I'm seeing, none of the core JS files have been changed.  package.json appears to be valid based on the relevant checks from the Makefile:

```js
$ eslint --ext .js,.njs,.json,.html,.htm package.json bower.json
$ jshint --show-non-errors package.json bower.json
```

(btw: if you run `make help` you will see a list of the tasks)

DxCx:
hey @wolfgang42 i'm the mainter of ts-xlsx package,
I'm very happy to see this PR because this is the only reason i had to maintain it..
when this will be merged i'll add deprecation warning to it instructing people to just use this instead.

types/index.d.ts
```diff
@@ -0,0 +1,435 @@
...
+
+export interface ColInfo {
+    /**
+     * Excel's "Max Digit Width" unit, always integral
+     */
+    MDW?: number;
+    /**
+     * width in Excel's "Max Digit Width", width*256 is integral
+     */
+    width: number;
```
DMan577: As I've stated in the previous PR, width is an optional property.
Notice the comment about the priority order:
> 1) use width field if available 2) use wpx pixel width if available 3) use wch character count if available

Also, I've been using it with only `wch` and it works fine. So `width` is not required.

SheetJSDev: @DMan577 this was a mistake from the flow definitions, width is optional and we'll correct it after the merge.

SheetJSDev:
Thanks everyone!

@DxCx @DMan577 note we're going to change a few names:
- calling a cell address `Cell` is somewhat confusing, so that's going to become `CellAddress`
- WorkSheetCell suggests that the cell only applies to worksheets.  As we flesh out chart sheets etc, we'll want to split off a common cell object type that represents the baseline

If there are any other name changes that make sense, it's best to do them before pushing the next release.

PR #700, opened 2017-06-19
gets Sheet name for multi cell named range, fixes #680

### LINKED ISSUE #680: workbook.Names Ref returns **MISSING** sheet reference when named range refers to multiple cells
workbook.Names returns Ref: "\*\*MISSING\*\*!<CellRef>" when the named range refers to a multi-cell range. Single cell and dynamic named ranges work as expected (NamedRanges.xlsb is attached in the zip).

```
var wb = XLSX.readFile('NamedRanges.xlsb');
console.log(wb.Workbook.Names);
```

returns

`{"0":{"Name":"DynamicNamedRange","Comment":"","Ref":"OFFSET(Sheet1!$A$5,0,0,2,1)"},"1":{"Name":"MultiCellNamedRange","Comment":"","Ref":"**MISSING**!A2:A4"},"2":{"Name":"SingleCellNamedRange","Comment":"","Ref":"Sheet1!$A$1"}}`

[NamedRanges.zip](https://github.com/SheetJS/js-xlsx/files/1066467/NamedRanges.zip)

fzumstein:
Please have a look, this seems to resolve the issue for us.
Do you want us to add a test case or do you prefer to do that yourself?

### DIFF
--- a/bits/62_fxls.js
+++ b/bits/62_fxls.js
@@ -811,7 +811,7 @@ function stringify_formula(formula/*Array<any>*/, range, cell/*:any*/, supbooks,
 				break;
 			case 'PtgArea3d': /* 2.5.198.28 TODO */
 				type = f[1][0]; ixti = /*::Number(*/f[1][1]/*::)*/; r = f[1][2];
-				sname = (supbooks && supbooks[1] ? supbooks[1][ixti+1] : "**MISSING**");
+				sname = supbooks && supbooks[1] ? supbooks[1][ixti+1] : (supbooks.SheetNames[ixti] ? supbooks.SheetNames[ixti] : "**MISSING**");
 				stack.push(sname + "!" + encode_range((r/*:any*/)));
 				break;
 			case 'PtgAttrSum': /* 2.5.198.41 */


SheetJSDev:
This doesn't seem to work on all of the files from the test corpus.
After the line in question, add a line that throws an error if the field is missing:

```js
				sname = (supbooks && supbooks[1] ? supbooks[1][ixti+1] : supbooks.SheetNames[ixti]);
				if(!sname) throw new Error("missing name"); // <-- add this line
				stack.push(sname + "!" + encode_range((r/*:any*/)));
```

To check against the test corpus, you can run the following command:

```bash
$ env WTF=1 FMTS=xlsb make test
```

(The `WTF=1` part throws errors that would otherwise be caught and `FMTS=xlsb` restricts to reading only XLSB files.)

The first fail case is [`test_files/2013/apachepoi_50939.xls.xlsb`](https://github.com/SheetJS/test_files/blob/master/2013/apachepoi_50939.xls.xlsb?raw=true), tripping on the newly-added error.

fzumstein:
Hi, sorry for the late reply. I've fixed the commit to fall back again to `**MISSING**`.

It still fails somewhere with the `if(!sname) throw new Error("missing name");` line, but that is the same behaviour as on `master`, so I don't think that's being caused by my changes.

Let me know know what you think. Thanks!

fzumstein:
@SheetJSDev have you been able to look at my fixed commit? see previous comment. thanks for your feedback!

SheetJSDev:
I wasn't saying your commit wasn't useful for your case, but if we're going to change the function processing it's better to fixit in one fell swoop  :)

Looked into this a little bit, https://github.com/SheetJS/js-xlsx/issues/680#issuecomment-307869085 was my initial brain dump on the matter.

The XLSB workbook processor is missing a few records like:

```
0x0165: /* 'BrtSupSelf' */
0x0166: /* 'BrtSupSame' */
0x0163: /* 'BrtSupBookSrc' */
0x029B: /* 'BrtSupAddin' */
0x016A: /* 'BrtExternSheet' */
0x0169: /* 'BrtPlaceholderName' */
```

After adding some initial implementation, we ran into a few more random cases like https://github.com/SheetJS/test_files/blob/master/2013/apachepoi_ex45978-extraLinkTableSheets.xls.xlsb?raw=true .  I suspect the external links files need to be processed.

fzumstein:
OK thanks for the feedback!

reviewher:
The particular part was reworked, but you've been credited on a related commit involving incorrect quoting of #REF in an unknown 3d reference.

PR #720, opened 2017-07-06
Update SheetJS-vue.js - Changed from readAsBinaryString (deprecated) to readAsArrayBuffer()

avrtau:
As per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsBinaryString), the `readAsBinaryString()` is 

> deprecated as per the 12 July 2012

The SheetJS wiki has an [article](https://github.com/SheetJS/js-xlsx/wiki/Reading-XLSX-from-FileReader.readAsArrayBuffer()) on using `readAsArrayBuffer()` to read an XLSX file.
The code here is a copy of the code provided in the article with very minor changes.

### DIFF
--- a/demos/vue/SheetJS-vue.js
+++ b/demos/vue/SheetJS-vue.js
@@ -28,12 +28,28 @@ function s2ab(s) {
 Vue.component('html-preview', {
 	template: SJSTemplate,
 	methods: {
+		// as per: https://github.com/SheetJS/js-xlsx/wiki/Reading-XLSX-from-FileReader.readAsArrayBuffer()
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Thanks a lot!

MDN appears to be incorrect here.  According to https://w3c.github.io/FileAPI/#readAsBinaryString it's still valid:

> The use of readAsArrayBuffer() is preferred over readAsBinaryString(), which is provided for backwards compatibility.

Since the spec does recommend readAsArrayBuffer we'll accept the change.

Issue #719, opened 2017-07-06
Cells In CSV With % Get Converted to Dates


DZGoldman:
To reproduce, drag a CSV file with % symbols in some cells into the live demo http://oss.sheetjs.com/. Some, but not necessarily all, will get, converted to dates. For example, 

![screen shot 2017-07-06 at 12 10 57 pm](https://user-images.githubusercontent.com/13685235/27921051-88d0b106-6244-11e7-99cb-1fe6101f6b0e.png)

converts to:
![screen shot 2017-07-06 at 12 13 37 pm](https://user-images.githubusercontent.com/13685235/27921091-a7d0ee72-6244-11e7-9406-6cdbb64f9be8.png)

I'm not sure if any of this is expected/desired behavior, but ideally I'd like a way to override it entirely.

SheetJSDev:
Gotta love JS:

```js
> [1,2,3,4,5,6].map(x=>new Date(x + "0%"))
[ 2001-10-01T04:00:00.000Z,
  Invalid Date,
  Invalid Date,
  2040-01-01T05:00:00.000Z,
  1950-01-01T05:00:00.000Z,
  1960-01-01T05:00:00.000Z ]
```

There's currently no option to inhibit the parsing behavior but it's fairly straightforward to add.  We use `raw` in other contexts to indicate raw values, so it probably makes sense to stick to that name.

The relevant function is https://github.com/SheetJS/js-xlsx/blob/master/bits/40_harb.js#L513 and the simplest solution looks like:

```diff
-                       if(s.charCodeAt(0) == 0x3D) { cell.t = 'n'; cell.f = s.substr(1); }
+                       if(o.raw) { cell.t = 's'; cell.v = s; }
+                       else if(s.charCodeAt(0) == 0x3D) { cell.t = 'n'; cell.f = s.substr(1); }
```

DZGoldman:
Great! If it's that simple, the change would be much appreciated.

DZGoldman:
Thanks for the merge, but I'm actually still seeing the issue - I updated to version `0.10.8` in my project and it persists.

DZGoldman:
(Side note: It appears that the demos are still on version 0.10.2 - if the version is updated we can test the changes there without confusion.)

SheetJSDev:
You need to pass the `raw:true` option to suppress the value interpretation.  See https://runkit.com/sheetjs/5964eadbea82990011acd7f7 for an example

[...]

PR #722, opened 2017-07-09
Added a check for a raw string in .csv files

TMiller00:
Implemented the recommend solution to #719


SheetJSDev:
Thanks a lot!

dginev:
Just wanted to extend our thanks at @authorea for the fix and the great library, we're looking forward to using it extensively in production!

PR #735, opened 2017-07-16
Fix potential infinite loop condition in test suite

TeamworkGuy2:
I was doing some testing with js-xlsx's test suite and ran into a test that was hanging.  The problem was cmparr()'s for-loop condition.  Switching from `!=` to `<` fixed the hung test and caused it to error correctly. Thought I'd drop the fix here in-case it's useful to js-xlsx.

### DIFF
--- a/test.js
+++ b/test.js
@@ -683,7 +683,7 @@ function custprop(wb) {
 	assert.equal(wb.Custprops.Counter, -3.14);
 }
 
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Good catch! Out of curiosity, what platform triggered the error?

TeamworkGuy2:
I made some custom changes to the js-xlsx test suite in relation to a TypeScript project I was working that has some special excel import/export requirements. 
The issue doesn't occur in a clean clone of the js-xlsx repository, just related to some custom changes I made that happened to uncover this.

PR #745, opened 2017-07-22
Clarification about the SystemJS configuration for custom project

rcfrias:
installing the module with npm would always download the 'xlsx.full.min.js' in the same directory.


SheetJSDev:
This is only true if your angular build is set up that way.  Using SystemJS in the browser, the final location might be different.  (The ecosystem is so complex 😦 )

PR #765, opened 2017-08-05
Updated code style

paustint:
Minor stylistic tweaks and removed the use of the "scope" variable in favor of using fat arrow syntax to retain the context of "this" without the need for an additional variable.


SheetJSDev:
@paustint looks good!  Can you squash down to one commit with a message like: `Improved Code Style of Angular 2+ Demo [ci skip]` (no need to list the specific changes in the commit message)

Issue #768, opened 2017-08-08
tag regex doesn't handle attributes containing >


MonochromeChameleon:
I've recently had a number of excel files fail when parsing the styles, with a fairly unclear error message:

```
TypeError: First argument must be a string, Buffer, ArrayBuffer, Array, or array-like object.
at fromObject (buffer.js:262:9)
at Function.Buffer.from (buffer.js:101:10)
at Buffer (buffer.js:80:17)
at utf8readc (.xlsx/xlsx.js:1717:52)
at parse_numFmts (.xlsx/xlsx.js:7046:23)
at parse_sty_xml (./xlsx/xlsx.js:7147:34)
at parse_sty (./xlsx/xlsx.js:12836:9)
at parse_zip (./xlsx/xlsx.js:17104:26)
at read_zip (./xlsx/xlsx.js:17393:9)
at Object.readSync [as read] (./xlsx/xlsx.js:17418:68)
```

I've tracked the issue down to the fact that the current regex used for detecting tags is `/<[^>]*>/g`, which can't handle `<numFmt numFmtId="164" formatCode="[>0]General" />` as it matches the closing angle bracket inside the formatCode attribute and interprets that as the end of the tag. The regex should be updated to include any number of attributes, disregarding their content, prior to the end of the tag.

SheetJSDev:
@MonochromeChameleon thanks for the report!   Do you know how were these files produced?  In your example, if you put the format code in Excel and save, the generated xml has the encoded `&gt;`

MonochromeChameleon:
I'm not sure exactly where the format code was coming from - I wasn't adding it myself! I'm using excel on a mac instead of on windows, which may have made a difference?

PR #769, opened 2017-08-08
Update tag regex to handle angle brackets in attributes, fixes Issue #768

MonochromeChameleon:
This fixes issue 768 by replacing the tagregex with one which ignores the contents of attributes for parsing XML


SheetJSDev:
Looks good, but the tests are failing on a test file with strange whitespace:

```xml
<workbook
 xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main"
 xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships">
```

Your regex assumes there's one whitespace character after the tag name.  The fixed regex is 

```js
/<[\/\?]?[a-zA-Z0-9:]+(?:\s+[^"\s?>\/]+=(?:"[^"]*"|'[^']*'|[^'">\s]+))*\s?[\/\?]?>/g;
                           ^ added this plus
```

If you can fix the regex and amend the commit with a new message (keep the first line within 50 characters), we can merge as-is.  Here's an example:

```
new tag regex handles angle brackets in attributes

fixes #768
```

MonochromeChameleon:
Updated with that +, as well as one for the final \s outside the attribute capture

PR #784, opened 2017-08-17
Fix fuzzydate string matching logic

### LINKED ISSUE #772: Convert html to xlxs: numbers became date
We are converting html table to xlxs/csv, for some data, the number became date.
For example, $41.08 was exported as "31-07-2041  4:00:00 PM"
Different combination of options have be tested:

var wopts = { bookType: type, bookSST: false, type: 'binary', raw: true, cellNF: true, cellDates : false, dateNF: 2}, 

Any suggestion?

MonochromeChameleon:
I'm seeing a bug in the fuzzydate function when uploading a CSV which includes a cell with the value `THIS IS A BUG 90` - it gets parsed to 1 January 1990. It looks to me like the update in commit 5b67ac0 to fix #772 altered the logic as to what constituted a date, but should also have swapped the returned values to reflect that updated logic.


SheetJSDev:
Good catch!

PR #802, opened 2017-09-04
Spelling

jsoref:
You can of course drop commits.
The filename change is a bit odd -> SheetJS/jxls#2 (someone would have to fix this repo once that PR is taken, if it is).

The password change is because it's an annoying false positive. afaict, there's no particular requirement to misspell the word by omitting a character, and thus it's nicer for spelling tools if you just append a suffix which should also result in a non-matching password. There's also a family of tokens such as `C4tegory`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2111,7 +2111,7 @@ the metadata the output is valid HTML, although it does accept bare `&` symbols.
 
 `make test` will run the node-based tests.  By default it runs tests on files in
 every supported format.  To test a specific file type, set `FMTS` to the format
-you want to test.  Feature-specific tests are avaialble with `make test_misc`
+you want to test.  Feature-specific tests are available with `make test_misc`
 
 ```bash
 $ make test_misc   # run core tests
--- a/bits/18_cfb.js
+++ b/bits/18_cfb.js
@@ -381,7 +381,7 @@ var ENDOFCHAIN = -2;
 var HEADER_SIGNATURE = 'd0cf11e0a1b11ae1';
 var HEADER_CLSID = '00000000000000000000000000000000';
 var consts = {
-	/* 2.1 Compund File Sector Numbers and Types */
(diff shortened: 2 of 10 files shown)


SheetJSDev:
Thanks for looking into this!

The `dynamic` error corresponds to a test file with that name (it's pulled from the [jXLS project test suite](https://github.com/SheetJS/jxls/blob/master/jxls-examples/src/main/resources/templates/dynamicolumns.xls) ), so that should be preserved.

Instead of `passwor_`, change the password to `passwords` or `Password` or `PASSWORD` (pick your favorite).

Other weird names like `C4tegory` appear in the [test script](https://github.com/SheetJS/js-xlsx/blob/master/test.js#L1255) as fake names for the various properties.  The pertinent test starts from the JS representation, writes files, reads the new files, and verifies that the properties of the new file match the original representation.  The [object's keys](https://github.com/SheetJS/js-xlsx/blob/master/test.js#L1255-L1269) should not be changed, but the values can be changed.  Feel free to pick your favorite sesquipedalians!  

Please undo the commit corresponding to `dynamic`, change `passwor`, and squash down to one commit with the message `spelling [ci skip]` (the `[ci skip]` part will skip travis tests)

jsoref:
Oh, I just realized why I didn't see the files. I had deleted them to avoid spellchecking them...

SheetJSDev:
The other changes in #803 look good, if you can fold those changes from #803 into this PR with one commit, we can merge as-is! (normally we would do some git jujutsu but GH doesn't recognize a commit as a successful merge unless it has the same exact hash :/

Make the message for the combined commit:

```js
spelling [ci skip]

- fixed spelling errors in README and code (fixes #802)
- replaced garbled names in property tests (fixes #803)
```

(GH will automatically add a note in both PRs linking to the commit)

PR #871, opened 2017-11-03
CSV export can now be cleansed from new lines and trailing spaces

const-g:
Added an option `--clean` to make CSV export to remove new lines from cell content, and trim the content also

### DIFF
--- a/bin/xlsx.njs
+++ b/bin/xlsx.njs
@@ -19,29 +19,30 @@ program
 	.option('-B, --xlsb', 'emit XLSB to <sheetname> or <file>.xlsb')
 	.option('-M, --xlsm', 'emit XLSM to <sheetname> or <file>.xlsm')
 	.option('-X, --xlsx', 'emit XLSX to <sheetname> or <file>.xlsx')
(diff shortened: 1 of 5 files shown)


SheetJSDev:
@const-g overall makes sense, but there are a few comments:

1) should this also apply to other formats, or is this appropriate only for CSV / TXT?

2) https://docs.sheetjs.com/#delimiter-separated-output there are other options that are not exposed to the CLI, namely `strip` and `blankrows`. The "strip" option removes trailing delimiters (`foo,bar,,` -> `foo,bar`) and "blankrows" omits blank rows.  Should those also be set under `--clean`?

3) Your editor also made a bunch of changes to package.json :( can you revert them?

4) Can you undo all of the whitespace changes to bits/90_utils.js and re-make?

5) We'd prefer the commit one-line summary to be within 50 characters.  Can you change the message to something like:

```
Plaintext export "clean" option and CLI "--clean"

Clean removes newlines and trims 
```

SheetJSDev:
If you can explain why this is interesting/useful we will reconsider merging.

PR #886, opened 2017-11-15
correct typo

mzvast:
add the missing comma in data array


SheetJSDev:
Good catch!  Can you also make the same change to the react example in https://github.com/SheetJS/js-xlsx/blob/master/demos/react/README.md and squash to one commit with the message:

```
correct vue and react README typos [ci skip]
```

PR #918, opened 2017-12-14
Add columns with empty headers to JSON output

vjau:
With current sheet_to_json function, when using first row as data headers, when a column's first cell is empty, the whole column is not included in the result.
With this small PR, columns with empty header are included in the result, with the special "__EMPTY" key.
If there are multiple empty columns, they get the numbered "__EMPTY", "__EMPTY_1", "__EMPTY_2", etc...

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -18815,7 +18815,7 @@ function sheet_to_json(sheet, opts) {
 			case 2: hdr[C] = cols[C]; break;
 			case 3: hdr[C] = o.header[C - r.s.c]; break;
 			default:
-				if(val == null) continue;
+				if(val == null) val = {w: "__EMPTY", t: "s"};
 				vv = v = format_cell(val, null, o);
 				counter = 0;
 				for(CC = 0; CC < hdr.length; ++CC) if(hdr[CC] == vv) vv = v + "_" + (++counter);


SheetJSDev:
@vjau makes sense, but should those missing values be enumerable?  For example, the default output includes the [non-enumerable `__rowNum__` property](https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L47-L48) for every row

Also, can you amend the commit message so that the first line is within 50 characters, like:

```
sheet_to_json: __EMPTY keys for missing headers

closes #864
```

vjau:
Thank you for taking the time to review this PR.
I have updated the commit message per your request.
Why do you think those values should be non enumerable. Actually, it's not the values that are empty but only the header name. With the current code (before the PR), the whole data of the column with missing header name is dropped, this can cause to critical (because silent) loss of data .

SheetJSDev:
I agree that the missing data should be present somehow.  To explain the enumerability question, consider the sheet with a few missing column headers:

| |A|B|C|D|E|
|-|-|-|-|-|-|
|**1**| |a|b| |c|
|**2**|1|2|3| |5|
|**3**| |3|4|5|6|

As it currently stands, with `{raw:true}` you get

```js
var data_old = [
    {a:2, b:3, c:5},
    {a:3, b:4, c:6}
]
```

Setting the [option `header:1` will return an array of arrays](https://docs.sheetjs.com/#json), which will never drop cells.  With `{raw:true, header:1}` you get every cell:

```js
var aoa_old = [
    [     , "a" , "b" ,     , "c" ], // <-- the gaps are array holes
    [  1  ,  2  ,  3  ,     ,  5  ]
    [     ,  3  ,  4  ,  5  ,  6  ]
]
```

With your current proposal you get:

```js
var data_new = [
    {a:2, b:3, c:5, __EMPTY   :1 },
    {a:3, b:4, c:6, __EMPTY_1 :5 }
]
```

The non-enumerable version looks like:

```js
var data_ne = [ { a: 2, b: 3, c: 5 }, { a: 3, b: 4, c: 6 } ];
Object.defineProperty(data_ne[0], '__EMPTY',   { enumerable: false, value: 1 });
Object.defineProperty(data_ne[1], '__EMPTY_1', { enumerable: false, value: 5 });

```

The enumerability question boils down to this: should the following code segment print out the `__EMPTY` entries?

```js
data.forEach((row) => {
    for(let [k,v] of Object.entries(row)) console.log(k,v);
});
```

To see the difference, try it out:

```js
/* current code */
> data_new.forEach((row) => { for(let [k,v] of Object.entries(row)) console.log(k,v); });
a 2
b 3
c 5
__EMPTY 1
a 3
b 4
c 6
__EMPTY_1 5

/* non-enumerable version */
> data_ne.forEach((row) => { for(let [k,v] of Object.entries(row)) console.log(k,v); });
a 2
b 3
c 5
a 3
b 4
c 6
```

There is precedent for non-enumerable properties: [each row object has `__rowNum__` as a non-enumerable property](https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L46-L48) to preserve the original row number

vjau:
It seems to me that __rownum__ could be considered metadata and thus should not be enumerable. 
But i think the contents of the cells in columns without headers are real data.
Is it the generated key name that causes you a problem ?
As you know for...of has been implemented to make object traversal more idiot-proof without the usual checks, and IMHO the default behaviour should be to spit out all of the data.
Perhaps the __EMPTY name is a bad choice, i choose it to prevent name clash.
I discovered that you can name a key "undefined", perhaps undefined, undefined_1, undefined_2 would be a better choice ?

SheetJSDev:
It's easy to add features but difficult to change features once they are added, so a few moments of thought now spare us from future pain :)

I also added some tests and updated the bits in commit 028d7010a516383cb9a2fdd0f0a919392e77600a (the new commit used the author and date from this PR, so you should be marked as a contributor to the repo)

PR #937, opened 2018-01-02
bug fix: when the arrayBuffer is to large, __toBuffer will cause error

hacker1117:
The __toBuufer function use "x.push.apply",but the argument has limits. when the length of array is larger than 65536, it will cause an error: Maximum call stack size exceeded.(refer to: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply)
so i apply chunks of the array at a time.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -2255,7 +2255,7 @@ function write_double_le(b, v, idx) {
 	b[idx + 7] = (e >> 4) | bs;
 }
 
-var __toBuffer = function(bufs) { var x = []; for(var i = 0; i < bufs[0].length; ++i) { x.push.apply(x, bufs[0][i]); } return x; };
+var __toBuffer = function(bufs) { var x = [],QUANTUM = 32768; for(var i = 0; i < bufs[0].length; ++i) { for(var j=0,len= bufs[0][i].length; j<len;j+=QUANTUM) { buf_arr = bufs[0][i].slice(j,Math.min(j+QUANTUM,len));x.push.apply(x, buf_arr); }} return x; };
 var ___toBuffer = __toBuffer;
 var __utf16le = function(b,s,e) { var ss=[]; for(var i=s; i<e; i+=2) ss.push(String.fromCharCode(__readUInt16LE(b,i))); return ss.join("").replace(chr0,''); };
 var ___utf16le = __utf16le;


SheetJSDev:
On the read side, before adding the array input type, a helper function used to chunk in 10KB units to circumvent the same problem. Not sure why a similar change wasn't also applied here. 
 https://github.com/SheetJS/js-xlsx/blob/a8736580a58011d112032f656a0e21779f78aa2f/index.html#L91-L96

```js
function fixdata(data) {
	var o = "", l = 0, w = 10240;
	for(; l<data.byteLength/w; ++l) o+=String.fromCharCode.apply(null,new Uint8Array(data.slice(l*w,l*w+w)));
	o+=String.fromCharCode.apply(null, new Uint8Array(data.slice(l*w)));
	return o;
}
```

Here's a test fiddle that fails in the current version https://jsfiddle.net/pg6oghLu/ and passes with the proposed changes https://jsfiddle.net/pg6oghLu/1/ :

```js
// <pre id="results"></pre>
var verbose = false;
results.innerText = PRINTJ.sprintf("BOOKTYPE  TYPE   RESULT\n");
var wb = XLSX.utils.book_new();

/* Generate 500 rows of 20 columns each */
var data = [];
data[0] = ["Row Number"];
for(var j = 0; j < 19; ++j) data[0].push("Column " + j+1);

for(var i = 0; i < 499; ++i) {
  var o = ["Row " + i];
  for(j = 0; j < 19; ++j) o.push(i + j);
  data.push(o);
}

var ws = XLSX.utils.aoa_to_sheet(data);
XLSX.utils.book_append_sheet(wb, ws, "Data");

["xlsx", "xlsb", "biff8", "biff5", "biff2", "xlml", "ods", "fods", "csv", "txt", "sylk", "html", "dif", "dbf", "rtf", "prn", "eth"].forEach(function(btype) {
  ["binary", "array", "base64"].forEach(function(type) {
    var res = "Y";
    try { var out = XLSX.write(wb, {bookType:btype, type:type}); XLSX.read(out, {type:type}); } catch(e) { console.log(btype, type, e.message); res = "N"; }
    if(verbose || res != "Y") results.innerText += PRINTJ.sprintf("%9s %6s %s\n", btype, type, res);    	
  });
});
```

hacker1117:
I can't find where fixdata function in 'index.html'.So the problem is you have already has fixdata() function somewhere doing the same work, but it doesn't work? Or something else? 
I want to know what should i do for the next to fix this problem

hacker1117:
So should I apply this change to the fixdata() function? but i can't find this function in 'index.html'.:(

SheetJSDev:
@hacker1117 sorry I wasn't clear.  For the same reason that you encountered, there used to be a chunking algorithm in processing ArrayBuffers that used `QUANTUM=10240`  instead of the original 32768, assumably because of browser compatibility.  We'll rework the commit and add a test :)

SheetJSDev:
@hacker1117 The 10240 threshold works in IE8+ as evidenced in the browser test https://saucelabs.com/beta/builds/3de647aa4e09454fa30894170668a8d4 .  For whatever reason, IE8 was failing on 32768, which is probably why we used the lower number in the first place.

PR #1027, opened 2018-03-07
Updated types list, fixes #1026

### LINKED ISSUE #1026: incomplete ParsingOptions/WritingOptions interface in Typescript in index.d.ts
When trying to read an excel xml file generated in ag-grid, I stumbled upon encoding problem. 
Then I noticed, that the problem is fixed, when I read workbook with options  { type: 'string' } in browser console. 

However typescript is not letting me to build such a code, because types description clearly states only following allowed options:

```
 /** Options for read and readFile */
 export interface ParsingOptions extends CommonOptions {
     /** Input data encoding */
   type?: 'base64' | 'binary' | 'buffer' | 'file' | 'array';
```

Here is the description for the options:

https://www.npmjs.com/package/xlsx#input-type

The same goes for both WritingOptions options as well as ParsingOptions options

andrisarkameru:
Updated types options  "string" was missing, but otherwise functional.  Typescript would not let me build otherwise. https://www.npmjs.com/package/xlsx#input-type

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -110,7 +110,7 @@ export interface DateNFOption {
 /** Options for read and readFile */
 export interface ParsingOptions extends CommonOptions {
     /** Input data encoding */
-    type?: 'base64' | 'binary' | 'buffer' | 'file' | 'array';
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Good catch!

PR #1028, opened 2018-03-07
Module name should be consistently used in examples

georgiebin:
Module name in the readFile example code should be identical to the one used with the writeFile example and a prior example where XLSX is defined as the object the require('xlsx') returns


SheetJSDev:
Good catch! will merge then amend the commit message

Issue #1041, opened 2018-03-19
brtRevisionPtr breaks parsing


berzniz:
Parsing an `xlsb` file with `brtRevisionPtr` breaks with error:
```
Error: Unexpected record 3073 BrtRevisionPtr
```

According to https://msdn.microsoft.com/en-us/library/mt793316(v=office.12).aspx - this field should be ignored. When ignoring it (I monkey patched the final javascript file), it works and parses the file correctly.

I can not send the xlsb test file since it has sensitive information. I'm trying to create a blank test file that can be shared.

SheetJSDev:
Part of why it errors on unrecognized records is to find files with these records! Even if you can't publicly share it, we'd love a small test file (you can email dev@sheetjs.com or sheetjs@gmail.com)

PR #1042, opened 2018-03-19
Fixes xlsb file parsing with BrtRevisionPtr

berzniz:
This addresses #1041 (BrtRevisionPtr breaks parsing) by ignoring it.


SheetJSDev:
Will merge then amend the commit to include the issue number in the commit message.

For future reference: For GH to link the issue to the commit, you need to write "fixes #1041" in the body of the commit message

Issue #1017, opened 2018-03-04
JSZip 3.0 error


seanmars:
I need use JSZip for DataTable but it will issue the follow error

```
Uncaught Error: The constructor with parameters has been removed in JSZip 3.0, please check the upgrade guide.
```

seanmars:
Thanks for fix it!

Issue #869, opened 2017-11-02
js-xlsx not support jszip 3.x


hstarorg:
In `https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L1920`, there used the global JsZip lib at first. In fact, the `xlsx.js` all packed the jszip 2.x, so I think use global jszip is unnecessary.

In my project, I used the JsZip as global variable, so the xlsx has a compatibility question.(not support jszip 3.x)

hstarorg:
Though the dist `xlsx.full.min.js` ships with a version of JSZip, the xlsx have priority use the global JsZip.
See code : [https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L1920](https://github.com/SheetJS/js-xlsx/blob/master/xlsx.js#L1920)

martinnov92:
Can you reopen this issue please? Even when using `xlsx.full.min.js` it is not working

hstarorg:
@martinnov92 In my project, I removed the code that use global JsZip.

martinnov92:
@hstarorg This is not really solution in our project, we have a large code base :/ This should be fixed in xlsx in my opinion

hstarorg:
@martinnov92 But I think it's not quickly.

Issue #600, opened 2017-03-23
zip library


reviewher:
jszip stopped supporting the sync API, is it worth investing in a new JS ZIP library?

twogood:
Or make js-xlsx async?

Issue #408, opened 2016-05-03
I can't upgrade JSZip to ^3.0.0


limouri:
I can' t use the current version of **JSZip ^3.0.0**, i get the following log while i try to read an .xlsx file:   '**Error: The constructor with parameters has been removed in JSZip 3.0, please check the upgrade guide.(…)**'

jacobmischka:
JSZip 3.0.0 changed its api to make calls asynchronous. It would be impossible to use JSZip 3.0.0 in this package without essentially rewriting everything to use Promises. Not worth it.

SheetJSDev:
JSZip API changed and they completely removed the sync interface.  It looks like [jszip-sync](https://www.npmjs.com/package/jszip-sync) [source](https://github.com/ericvergnaud/jszip) supports sync, so that is definitely an option moving forward

SheetJSDev:
@limouri @jacobmischka for now to avoid problems, the jszip external dependency was removed and replaced with a local copy.  The bower points to a script that combines the jszip dependency and the core library, so there should be no need to use the external dependency.

dustincjensen:
I would vote for this to be re-opened.

This is, in my opinion a pretty massive flaw. If anyone does what I did and implement JSZip 3.0 to make a zip of images, then bring in js-xlsx, then their JSZip stops working.

It should be at least mentioned as incompatible in the readme, so people know they can or cannot use js-xlsx before spending time implementing it.

lyatziv:
I agree with @dustincjensen this is an issue that needs to be resolved. js-xlsx should either adopt a dependency that supports sync or just include it in the main source and not flag it as a dependency.
That would introduce bloat to users of the JSZip but ... ¯\_(ツ)_/¯ 
The other option is to re-write to support promises.

PR #1055, opened 2018-03-28
Upgrade to new jszip

furstenheim:
jszip v3 has async functions that return promises instead of blocking the thread.

I've adapted the code and all the tests to run with the new version.

Of course there would be some missing parts: documentation is not updated (now xlsx.read and xlsx.write return promises)
It won't work in older browsers as I'm heavily using async await, probably babel could be used. For my purposes, which is using the library in Node 8, it is enough, but maybe you are interested in the changes.

cheers


furstenheim:
This pr is related to a couple of issues #600 #1017 #869 #408

SheetJSDev:
@furstenheim to be clear, promises are "deferred".  When the actual work is being done, it will still block the main thread (pure JS is still single-threaded, not too dissimilar to Python).  Promises don't magically make things run off the main thread -- they merely shift the timing.  Platform APIs have to provide the functionality in a way that doesn't block the main thread, like the `zlib` bindings in nodejs.  The situation is similar to how NumPy circumvents the Python global interpreter lock. 

To see this, try running the following command in your node shell:

```js
setTimeout(() => console.log("done"), 1000); while(1);
```

setTimeout and most promise APIs don't preempt the execution.


If you want to process off the main thread in nodejs, check [the `server` demo for a few strategies](https://github.com/SheetJS/js-xlsx/tree/master/demos/server).  You can use `child_process.fork` with a simple JS script or use `exec` to call the `xlsx` bin script or use [the `tiny-worker` Web Worker "shim"](https://www.npmjs.com/package/tiny-worker)


Note: If you want to process off the main thread in web browser, the standard technique is to use a Web Worker.

aivaras-b:
@SheetJSDev it does make things "magical" because by deferring heavy calculation/work you are not blocking higher priority events which results in smoother experience at least for browser users.

PR #1077, opened 2018-04-17
add px unit to font-size

drewsmith:
Chrome marks font-size as invalid without a font size unit.

font_size_invalid


SheetJSDev:
The quoted units are `pt`, not `px`.  We will create a new commit and use the original commit info

PR #1122, opened 2018-05-27
Added an option to ignore hidden rows and cells while parsing a DOM table

### LINKED ISSUE #1115: Feature request: ignore hidden cells in table_to_sheet/book
It would be great if SheetJS ignores hidden rows and cells when parsing a DOM table using the `table_to_sheet` and `table_to_book` functions.

For example, the following table:

```html
<table>
  <tr>
    <td>One</td>
    <td style="display: none;">Two</td>
    <td>Three</td>
  </tr>
  <tr>
    <td>1</td>
    <td class="hidden">2</td>
    <td>3</td>
  </tr>
  <tr style="display: none;">
    <td>Раз</td>
    <td>Два</td>
    <td>Три</td>
  </tr>
</table>

<style>
  .hidden {
    display: 'none';
  }
</style>
```

Would be parsed as:

| One | Three |
|-----|-----|
| 1 | 3 |

Detecting whether the cell is hidden can be done by detecting a `display: none` style:

```js
if (getComputedStyle(cell).getPropertyValue('display') === 'none') {
    continue;
}
```

Ignoring hidden cells and rows can be optional:

```js
XLSX.utils.table_to_sheet(table, {
    ignoreHiddenRows: true,
    ignoreHiddenCells: true
});
```

If you support this feature, I can make a pull request.

Finesse:
Resolves #1115 

There is a new option called `display` in the `table_to_sheet` and `table_to_book` methods. If its value is `true`, the hidden `td`s and `th`s are not parsed. The default value of this option is `false`. Also hidden `tr`s become hidden in a parsed table when the `display` option is `false`.

Example. Suppose there is a table on a web page:

<details>
<summary>Table HTML</summary>

```html
<table>
  <tr>
    <td>One</td>
    <td style="display: none;">Two</td>
    <td>Three</td>
  </tr>
  <tr>
    <td>1</td>
    <td class="hidden">2</td>
    <td>3</td>
  </tr>
  <tr style="display: none;">
    <td>Раз</td>
    <td>Два</td>
    <td>Три</td>
  </tr>
</table>

<style>
  .hidden {
    display: 'none';
  }
</style>
```
</details>

If `display` value is `false`, the parsed table looks this way:

image

If `display` value is `true`, the parsed table looks this way:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -2055,6 +2055,7 @@ Both functions accept options arguments:
 |`dateNF`     |  FMT 14  | Use specified date format in string output          |
 |`cellDates`  |  false   | Store dates as type `d` (default is `n`)            |
 |`sheetRows`  |    0     | If >0, read the first `sheetRows` rows of the table |
+|`display`    |  false   | If true, hidden rows and cells will not be parsed   |
 
 
 <details>
--- a/bits/79_html.js
+++ b/bits/79_html.js
@@ -119,15 +119,22 @@ function parse_dom_table(table/*:HTMLElement*/, _opts/*:?any*/)/*:Worksheet*/ {
 	if(DENSE != null) opts.dense = DENSE;
 	var ws/*:Worksheet*/ = opts.dense ? ([]/*:any*/) : ({}/*:any*/);
 	var rows/*:HTMLCollection<HTMLTableRowElement>*/ = table.getElementsByTagName('tr');
-	var sheetRows = Math.min(opts.sheetRows||10000000, rows.length);
-	var range/*:Range*/ = {s:{r:0,c:0},e:{r:sheetRows - 1,c:0}};
+	var sheetRows = opts.sheetRows || 10000000;
+	var range/*:Range*/ = {s:{r:0,c:0},e:{r:0,c:0}};
 	var merges/*:Array<Range>*/ = [], midx = 0;
-	var R = 0, _C = 0, C = 0, RS = 0, CS = 0;
-	for(; R < sheetRows; ++R) {
-		var row/*:HTMLTableRowElement*/ = rows[R];
+	var rowinfo/*:Array<RowInfo>*/ = [];
+	var _R = 0, R = 0, _C, C, RS, CS;
+	for(; _R < rows.length && R < sheetRows; ++_R) {
+		var row/*:HTMLTableRowElement*/ = rows[_R];
+		if (is_dom_element_hidden(row)) {
+			if (opts.display) continue;
+			rowinfo[R] = {hidden: true};
+		}
 		var elts/*:HTMLCollection<HTMLTableCellElement>*/ = (row.children/*:any*/);
 		for(_C = C = 0; _C < elts.length; ++_C) {
-			var elt/*:HTMLTableCellElement*/ = elts[_C], v = htmldecode(elts[_C].innerHTML);
+			var elt/*:HTMLTableCellElement*/ = elts[_C];
+			if (opts.display && is_dom_element_hidden(elt)) continue;
+			var v/*:string*/ = htmldecode(elt.innerHTML);
 			for(midx = 0; midx < merges.length; ++midx) {
 				var m/*:Range*/ = merges[midx];
 				if(m.s.c == C && m.s.r <= R && R <= m.e.r) { C = m.e.c+1; midx = -1; }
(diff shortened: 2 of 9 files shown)


Finesse:
I have a question. I've changed the source files in the `bits` directory, built them by running `make clean && make` and successfully run the Mocha tests through console. Now I want to run browser tests but I can't find out how to build `tests/xlsx.core.min.js` and `tests/xlsx.full.min.js`. Must I build them? How to build them?

Finesse:
@SheetJSDev By the way, I noticed 6 errors related to dates while I run the tests. For example:

image

My timezone is UTC+10.

SheetJSDev:
Running the node tests is straight forward (make + make test).  If you install `jsdom` locally, it will use it to run the dom-related tests.

To run the browser tests against your changed version, `make dist` will generate the browser files (do not commit the dist directory) and `make ctestserv` will start a local server in the test directory on port 8000).  We run all of the tests through sauce labs (generating the browser matrix) when we're ready to do a version bump, but running in latest chrome is sufficient for now.

FYI Both of the new tests are failing in chrome:



There's a chance that getComputedStyle depends on the actual element being attached to the document.  Was this working in your tests?

The TZ tests probably have to be revisited, since [we currently don't have a test for Australia or regions like Guam](https://github.com/SheetJS/js-xlsx/blob/master/.travis.yml#L10-L36).  `Australia/Canberra` is +10:00 / +11:00 and `Pacific/Guam` is +10:00 but does not follow DST.  That's obvious an unrelated issue

Finesse:
@SheetJSDev 

> FYI Both of the new tests are failing in chrome

I saw this errors. They occured in my browser because I hadn't built `tests/xlsx.core.min.js` and `tests/xlsx.full.min.js`. Therefore I asked how to build them.

I built the distributive files and found out that `getComputedStyle` doesn't work when the DOM is not inserted to a document (even with inline CSS). As I can see in the test code, HTML doesn't get appended:

```js
var domelt = document.createElement('div');
domelt.innerHTML = html;
return domelt;
```

To solve this I did the following:

1. In the source code: the `element.style.display` fallback is used when `getComputedStyle(element).getPropertyValue('display')` returns empty string.
2. In the test code: `get_dom_element` appends the given element to the document and removes it after a test case is run. This makes a `<style>...</style>` code from the HTML be applied to the tables.

Now the browser tests pass.

Finesse:
Semaphore failed the task on one of the Nodes. It doesn't look like my code caused it. Could you please restart the task?

SheetJSDev:
Semaphore has been flaky the last few months.  It was incidentally added because Travis was flaky.  Both are testing various versions of node, and the Travis tests are passing, so don't worry about those failures.

We ran the browser tests against sauce labs and it passed: https://saucelabs.com/beta/builds/d1ec989664244c10aa6a1e81af764646

The PR is almost ready to merge! Two last items:

1) add a comment `/* global getComputedStyle */` just before the `get_get_computed_style_function` definition (unfortunately linters don't understand the implications of `typeof`, so you need to give a hint through the comment):

2) squash down to one commit, with a message like

```
DOM TABLE parse element visibility 

- Hidden rows are marked as hidden by default (mimics Excel behavior).
- `display` option for `table_to_*` replicates table as it appears.

Fixes #1115
```

Finesse:
@SheetJSDev I've squashed the commits to a single commit. Have I added the `/* global getComputedStyle */` comment properly?

SheetJSDev:
This is awesome @Finesse !

Finesse:
@SheetJSDev Thank you for merging my changes!

While I was working with this project, a faced some difficulties which are not associated with the source code complexity:

1. There is no main guide or checklist about what to do if I want to make a PR. It could have a note about what checks one needs to perform before making a PR and how to perform them.
2. NPM dev dependencies are used incorrectly. If I run `npm install`, they are installed locally. But the makefile uses them as if they are installed globally which causes errors. I had to detect and globally install the dependencies by myself. To fix it you can set the makefile up to call them by local paths, for example:
```diff
.PHONY: test mocha
test mocha: test.js ## Run test suite
-    mocha -R spec -t 30000
+    ./node_modules/.bin/mocha -R spec -t 30000
```
3. There are a lot of unfamiliar type comments in the code, e.g. `parse_dom_table(table/*:HTMLElement*/, _opts/*:?any*/)/*:Worksheet*/`. I didn't find a note about what they are and how to check them.

SheetJSDev:
1\. We'll add a contributing note.  Since we had that problem with testing the browser build using the dist version, we'll add a new test page that doesn't rely on the dist version.

For node, `make test_misc` will run a light version of the node-based tests (takes ~6 seconds).  For the browser, `make ctestserv` runs a local web server and you can access http://localhost:8000/ to start the tests.  `make lint` runs eslint and `make old-lint` runs jshint and jscs.  

On our end, depending on the feature in question, we may run the browser based test suite using the Sauce Labs cross-browser testing service or check test files manually against specific versions of Excel.

2\. Technically we should use `npx`, but that's only supported in modern NPM versions.  At a later point we will review which commands can be run locally, but the global ones will be documented in the contributing notes.

3\. Those are flow annotations.  The command is in the [Makefile](https://github.com/SheetJS/js-xlsx/blob/master/Makefile#L195)

```
$ flow check --all --show-all-errors --include-warnings
```

We don't consider any PR to be a "finished product".  Quibbles like lint errors or flaky CI tests will be resolved when we push the next version.

Finesse:
@SheetJSDev I think using global dependencies is not a good practice because different projects require different versions of packages, and if everyone follows this practice, we will need to reinstall the global dependencies every time we switch from project to project.

I didn't know that Flow annotations can be used with comments, I saw only [examples with the regular annotation](https://flow.org/en/docs/types/).

P.S. I don't try to change the project contribution style, I just give a feedback.

SheetJSDev:
https://flow.org/en/docs/types/comments/ the comment annotations are really neat because the code is still 100% valid JS.

Feedback appreciated :) We do things a certain way based on what worked at the time.  Every once in a while we should revisit those decisions.

KonstantinsS:
Hi.
When this feature will be available?

Issue #1168, opened 2018-07-02
Using cell objects in json_to_sheet?


mattandrews:
Thanks for this great library and all your hard work – it's really impressive.

I'm trying to export some database records into Excel. I've used `json_to_sheet()` to output something basic, but I wanted to add a `=HYPERLINK()` formula for one cell.

Reading the docs, I can see that I need a `.f` or `.l` key on the relevant cell object to use a formula. This doesn't work when I use `json_to_sheet()` though, eg:

```
let myArray = [
  {
    name: 'bob',
    age: 23,
    homepage: {
      f: '=HYPERLINK("http://www.google.com", "Google")'
    }
  },
  {
    name: 'alice',
    age: 25,
    homepage: {
      l: {
        '.Target': 'http://www.google.com',
        '.Tooltip': 'Google'
      }
    }
  },
];

let worksheet = XLSX.utils.json_to_sheet(myArray);
let workbook = XLSX.utils.book_new();
XLSX.utils.book_append_sheet(workbook, worksheet, "SheetJS");
```

In the outputted sheet, both rows have a blank "homepage" column (the others are fine).

I guess this means the import functions don't allow cell objects. Is my only option here to import the JSON and then traverse the sheet and find my homepage values and manipulate the cell objects from there? It would be great if I could just set these at import time (as outlined above) if at all possible!

SheetJSDev:
As it is currently shaped, cell objects are not passed through.  Fortunately its a simple fix!

https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L199-L207 is the block that builds the cell to add.  It could be changed as follows:

```js
			if(v && typeof v == 'object' && !(v instanceof Date)){
				ws[encode_cell({c:_C + C,r:_R + R + offset})] = v;
			} else {
				if(typeof v == 'number') t = 'n';
				else if(typeof v == 'boolean') t = 'b';
				else if(typeof v == 'string') t = 's';
				else if(v instanceof Date) {
					t = 'd';
					if(!o.cellDates) { t = 'n'; v = datenum(v); }
					z = o.dateNF || SSF._table[14];
				}
				ws[encode_cell({c:_C + C,r:_R + R + offset})] = cell = ({t:t, v:v}/*:any*/);
				if(z) cell.z = z;
			}
```

You'd still have to make an actual cell object (with `t:'n'`), but that's straightforward.

We'll make that change in the next release (or you can send a PR :)

[...]

PR #1169, opened 2018-07-03
Allow cell objects to be passed in when importing data from JSON

mattandrews:
This relates to https://github.com/SheetJS/js-xlsx/issues/1168 – it basically implements the fix that @SheetJSDev suggests, which I've confirmed works on my test case in the code listed there.

Essentially this allows us to pass a structured cell object into utilities like `json_to_sheet()`, eg:

```
let myArray = [
  {
    name: 'bob',
    age: 23,
    link: {
        f: 'HYPERLINK("http://www.google.com", "Google")',
        t: 'n',
        v: 'http://www.google.com'
    }
  }
];

let worksheet = XLSX.utils.json_to_sheet(myArray);
let workbook = XLSX.utils.book_new();
XLSX.utils.book_append_sheet(workbook, worksheet, "SheetJS");
```

Note: I couldn't get the indentation to work properly for this PR – I tried in two different editors and despite telling both to use spaces, not tabs, the output of this PR is different from how the code looks in both IntelliJ and VS Code:

image

Is there a required editor config to make this work?

### DIFF
--- a/bits/90_utils.js
+++ b/bits/90_utils.js
@@ -196,16 +196,20 @@ function sheet_add_json(_ws/*:?Worksheet*/, js/*:Array<any>*/, opts)/*:Worksheet
 			var v = JS[k];
 			var t = 'z';
 			var z = "";
-			if(typeof v == 'number') t = 'n';
-			else if(typeof v == 'boolean') t = 'b';
-			else if(typeof v == 'string') t = 's';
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Switch to vim :D but it honestly doesn't matter, we will fix it later

PR #1199, opened 2018-07-30
Update README.md

jimmywarting:
fetch can return a arraybuffer...

### DIFF
--- a/demos/xhr/README.md
+++ b/demos/xhr/README.md
@@ -116,27 +116,18 @@ axios("/upload", {method: "POST", data: fd});
 
 ## fetch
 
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Can you also update the actual [`fetch` example page](https://github.com/SheetJS/js-xlsx/blob/master/demos/xhr/fetch.html#L42) and preferably squash to a commit with a message like "update XHR demo"

SheetJSDev:
Rewrote the commit in d01e10c3f21482c87a42b56dc02e215f9ee78c30 but credited it to the same email / tz from the PR (so GH should show you as a contributor)

PR #1200, opened 2018-08-01
added start script to electron demo

kami23:
when I run the application without this line 
 "scripts": {
    "start": "electron ."
  }, 
it doesn't work and it gives me an error saying that the start script is missing.

### DIFF
--- a/demos/electron/package.json
+++ b/demos/electron/package.json
@@ -3,6 +3,9 @@
 	"author": "sheetjs",
 	"version": "0.0.0",
 	"main": "main.js",
+        "scripts": {
+            "start": "electron ."
+         },
 	"dependencies": {
 		"electron": "~1.7.x",
 		"xlsx": "*"


SheetJSDev:
The command is buried in the [Makefile](https://github.com/SheetJS/js-xlsx/blob/master/demos/electron/Makefile#L11) but it probably makes more sense as npm scripts

```make
.PHONY: run
run:
	electron .
```

There's a lint command that probably should also be added to the package.json:

```make
.PHONY: lint
lint:
	eslint *.js
```

Can you also change the makefile so that it runs the npm script:

```make
.PHONY: lint
lint:
	npm run lint

.PHONY: run
run:
	npm start
```

Finally, can you change the commit message to indicate that this is for the electron demo, something like "added start script to electron demo"

reviewher:
It looks like 5b1440a186ec3ce45704b4fe2b87e82efc2f4c2d fixed the issue but wasn't properly tagged

PR #1276, opened 2018-09-19
Support cell object in the function 'sheet_add_aoa'

hellosean1025:
Support cell object , make js-xlsx easier to use.
```js
describe('support object cell', function() {
  var data = require('./').utils.sheet_add_aoa (null, [
    ['url', 'name', 'id'],
    [
      {
        l: {
          Target: 'https://123.com',
        },
        v: 'url',
        t: 's',
      },
      'tom',
      'xxx'
    ],
  ]);
  it('base', function(){
    assert.deepEqual(data.A2,  { l: { Target: 'https://123.com' },
    v: 'url',
    t: 's' })
  })
})


```


SheetJSDev:
We made a similar fix to [`sheet_add_json`](https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L208) but not to `sheet_add_aoa`.  There are some issues like whitespace but we can fix it up and push an amended commit.

hellosean1025:
Ok.

Issue #874, opened 2017-11-03
datenum() not accounting for DST properly


pmalouin:
Creating this issue as recommended in #832.

Here is a jsfiddle to witness the problem: https://jsfiddle.net/gtt2srcu/

I have some problems with the `datenum` function, not behaving correctly when the `v` is not in the same timezone offset as the current time (ex. current time has DST offset and `v` does not have DST offset). After playing around with the code, I believe the offsetting needs to be dependent on the `v` value instead of current time, like this:
```
var basedate = new Date(1899, 11, 30, 0, 0, 0);
var basetimestamp = basedate.getTime();
var baseoffset = basedate.getTimezoneOffset();
function datenum(v/*:Date*/, date1904/*:?boolean*/)/*:number*/ {
	var epoch = v.getTime();
	if(date1904) epoch -= 1462*24*60*60*1000;
	// Account for offset differences between the basedate and v...
	var adj = (v.getTimezoneOffset() - baseoffset) * 60000;
	return (epoch - basetimestamp - adj) / (24 * 60 * 60 * 1000);
}
```
This is because if `basedate` and `v` have a different offset, we need to compensate this effect by applying the difference on the numerical value provided to excel. The offset of the current time does not seem relevant.

It's possible that the `numdate` function suffers from the same problem, but I haven't digged through it.

SheetJSDev:
It's actually apparent from reading raw files.  Remove the `.zip` extension from the linked files:

[dates.xlsb.zip](https://github.com/SheetJS/js-xlsx/files/1441949/dates.xlsb.zip)
[dates.xlsx.zip](https://github.com/SheetJS/js-xlsx/files/1441950/dates.xlsx.zip)

```
$ node -pe "require('xlsx').readFile('dates.xlsx', {cellDates:true}).Sheets.Sheet1.A2"
{ t: 'd', v: 2017-02-01T04:00:00.000Z, w: '2017-02-01 00:00:00' }
$ node -pe "require('xlsx').readFile('dates.xlsb', {cellDates:true}).Sheets.Sheet1.A2"
{ t: 'd', v: 2017-02-01T05:00:00.000Z, w: '2017-02-01 00:00:00' }
```

[...]

Issue #1332, opened 2018-10-30
Converting dates (from JS) gets incorrect results if crossing DST boundary


lalomartins:
(Actually also with different timezones, but you state in the README you make a choice of not handling timezones at all, so let's focus on the DST issue.)

This is a little difficult to explain though. Let's say we ended Daylight Saving Time last Sunday. Let d1 be a date today, and d2 be a date exactly a week ago.

To test more easily, let's say we set `dateNF: 'yyyy-mm-dd"T"hh:MM:ss'`. Then put d1 and d2 in two columns in the same row and export a csv. (The bug is not actually related to csv, it's just easier to test if we convert to string.)

Expected: the time part (after T) is identical for both.

Actual: the time part for d2 will be off by one hour (e.g. 10:28 instead of 11:28).

The reason for this is [the dnthresh constant](https://github.com/SheetJS/js-xlsx/blob/9866dfc010338394e4cfcd33a9fbc15dae017ee5/bits/20_jsutils.js#L35) is calculated based on the timezone offset of the current time as of when the library is loaded. So dates with any other offset (including same timezone but different DST status) will break the calculation.

PR coming up in a few minutes.

SheetJSDev:
Thanks for looking into this @lalomartins !

The overarching problem is that Excel doesn't actually have a proper concept of time.  It's shoehorned into the numeric model in a very simplistic way, where each calendar day has 86400 equally-spaced seconds and midnight is pinned to 0.  This very clearly creates problems around DST.

lalomartins:
In our case it was an actual bug in our app (GrabCAD Print). Users noticed times on the CSV and XSLX exports were inconsistent with what they would see in the GUI and we tracked it down to DST shift. And the fix is relatively simple, so I got an OK from the company to contribute a PR.

cemremengu:
Spent hours trying to figure out why our tests were failing :smile:

Any ETA for the patch release?

guyko81:
I have issue with DST but in a more serious way: when importing pure datepart (no hours) from CSV the date jumps 1 full day when we move over DST boundary.
It's a rounding issue, but don't know where to fix in the code. 

Here is the actual date and the problem (date format | number format):

Expected:
2013.10.28 | 41575
2013.10.27 | 41574
2013.10.26 | 41573

Actual
2013.10.28 | 41575
2013.10.26 | 41573.95833
2013.10.25 | 41572.95833

[...]

PR #1333, opened 2018-10-30
Fix for #1332: Converting dates (from JS) gets incorrect results if crossing DST boundary

lalomartins:
Correctly convert dates that have a different UTC offset than the current local time. See #1332 for more details.

### DIFF
--- a/bits/20_jsutils.js
+++ b/bits/20_jsutils.js
@@ -49,7 +49,6 @@ function numdate(v/*:number*/)/*:Date*/ {
 	}
 	return out;
 }
-module.exports = {basedate, refdate, dnthresh, refoffset, datenum, numdate};
 
 /* ISO 8601 Duration */
 function parse_isodur(s) {
--- a/bits/20_jsutils.js
+++ b/bits/20_jsutils.js
@@ -32,17 +32,24 @@ function evert_arr(obj/*:any*/)/*:EvertArrType*/ {
 }
 
 var basedate = new Date(1899, 11, 30, 0, 0, 0); // 2209161600000
-var dnthresh = basedate.getTime() + (new Date().getTimezoneOffset() - basedate.getTimezoneOffset()) * 60000;
 function datenum(v/*:Date*/, date1904/*:?boolean*/)/*:number*/ {
 	var epoch = v.getTime();
 	if(date1904) epoch -= 1462*24*60*60*1000;
+	var dnthresh = basedate.getTime() + (v.getTimezoneOffset() - basedate.getTimezoneOffset()) * 60000;
 	return (epoch - dnthresh) / (24 * 60 * 60 * 1000);
 }
+var refdate = new Date();
+var dnthresh = basedate.getTime() + (refdate.getTimezoneOffset() - basedate.getTimezoneOffset()) * 60000;
+var refoffset = refdate.getTimezoneOffset();
 function numdate(v/*:number*/)/*:Date*/ {
 	var out = new Date();
 	out.setTime(v * 24 * 60 * 60 * 1000 + dnthresh);
+	if (out.getTimezoneOffset() !== refoffset) {
+		out.setTime(out.getTime() + (out.getTimezoneOffset() - refoffset) * 60000);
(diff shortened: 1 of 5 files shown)


lalomartins:
I'd love to include a test, but this is really finicky to reproduce, unfortunately; you need to be in a place that has DST, then compare a date with DST and one without. Even if there is a way to set or fake the local time and zone, that would probably be global and break all other tests.

lemmingworks:
Hi. I was running into a similar issue when using this package to parse csv files. Now that the UK is back to GMT, dates that were in BST (GMT+1) are being parsed as the day before because of the difference in UTC offset.

I've noticed that there are a couple of failing tests as a result of the changes in this PR:
`roundtrip features
       should preserve dates
         [n] -> (d) -> [d] -> (n):`

With regards to adding tests to replicate the scenario for this fix, have you had a look at the **timekeeper** npm module? It allows you to freeze time at a certain date, and then unset time again after your tests run. However, I have noticed that the xlsx package needs to be imported after the time has been frozen otherwise it will not work.

lemmingworks:
Hi @lalomartins 

I have fixed the failing units that I mentioned in my previous post, but I don't have permissions to push them to GrabCAD:master or to a branch.

It would be great if you could give me temporary permissions to push either to the GrabCAD:master, or to a branch where I could create a pull request.

The reason that the tests were failing was that, one of the existing tests uses the datenum() function, which you changed as part of your pull request, but as the datenum() function is not exposed as part of the xlsx api, someone had copy and pasted the old version into the test file, and this just needed to be updated to your new version.

This doesn't seem very nice, or maintainable going forward, but I'm just working with what's there.

lalomartins:
Well, thanks for pitching in to that!

I guess the normal github way to do it would be to fork and make a PR, right? I actually can't give you permissions on GrabCAD repo since it belongs to the company. You can do the “compare between forks” thing when creating the PR.

lalomartins:
If you don't want to bother with all that work you can also produce a patch file and put it somewhere I can get to.

lemmingworks:
@lalomartins No probs. I've submitted a new PR for this, with the fixed unit tests.

lalomartins:
I merged your changes as well, so now the two PRs are equivalent.

lemmingworks:
@lalomartins Great. I've closed my version to avoid confusion.

phani17c:
@SheetJSDev Could I know when this fix will be merged and make available?

mathieulj:
Unless im mistaken, this would also fix #874.

KrakenTyio:
Any news with this issue? i think should be merged

thilko:
Is the PR ready to merge? We would need that fix in production. Thx!

pierre-aurele-martin:
This should be merged ASAP :) Thanks !

coryasilva:
Is there an ETA on this merge?

dsbhough:
Please can someone supply an ETA on this merge?

Ugoku:
We're also running into this problem. Please merge this @SheetJSDev

SheetJSDev:
Thanks a lot @lalomartins and apologies for sitting on this, we've been hesitant to touch the date logic but finally decided to do it as part of the 0.16 release

chce:
What's the ETA on 0.16? @SheetJSDev

sahilsaid:
Did this get merged in? I am seeing this issue on latest (0.16.2)

Ugoku:
It's merged in the Pro version at least. For the Pro version, you have to specify `XLSX.set_date_style(1);` to get it to work.

SheetJSDev:
@sahilsaid this was merged in and the default was changed in this build.  In our Pro builds, the old date style is still the default, but you can change it using `XLSX.set_date_style` as mentioned by @Ugoku

sahilsaid:
Thanks @Ugoku and @SheetJSDev 
I am using the community version. Any idea when this fix will be available for community version?

christoph-hue:
so this really is only supported in pro version?? i'd rather implement the import with a different library then.

Issue #1368, opened 2018-11-28
Getting error: TypeError: Cannot read property 'replace' of undefined


vishenkov:
Hello!
Getting the error while reading this file.
This file is not damaged: opens correctly in Numbers (version 5.3 (5989)). Also correctly parsed by other libraries.

**demo repository (file included)**: https://github.com/vishenkov/xlsx-error-demo
**sandbox**: https://codesandbox.io/s/github/vishenkov/xlsx-error-demo
**file**: https://drive.google.com/file/d/1lSR5ZLxQkm0OYLxveYYPVXbpz9t5SExw/view?usp=sharing

```
/xlsx-error-demo/node_modules/xlsx/xlsx.js:10015
                comment.t = rt.t.replace(/\r\n/g,"\n").replace(/\r/g,"\n");
                                 ^

TypeError: Cannot read property 'replace' of undefined
    at /xlsx-error-demo/node_modules/xlsx/xlsx.js:10015:20
    at Array.forEach (<anonymous>)
    at parse_comments_xml (/xlsx-error-demo/node_modules/xlsx/xlsx.js:10003:65)
    at parse_cmnt (/xlsx-error-demo/node_modules/xlsx/xlsx.js:15094:9)
    at parse_comments (/xlsx-error-demo/node_modules/xlsx/xlsx.js:9945:16)
    at parse_zip (/xlsx-error-demo/node_modules/xlsx/xlsx.js:20021:19)
    at read_zip (/xlsx-error-demo/node_modules/xlsx/xlsx.js:20249:9)
    at readSync (/xlsx-error-demo/node_modules/xlsx/xlsx.js:20314:69)
    at Object.readFileSync (/xlsx-error-demo/node_modules/xlsx/xlsx.js:20328:9)
    at Object.<anonymous> (/xlsx-error-demo/index.js:3:23)
```

SheetJSDev:
The underlying issue has been resolved in version 0.15.0

raulgomezn:
Thanks

PR #1387, opened 2018-12-17
Validate the property to avoid an error

raulgomezn:
Fixed bug Cannot read property 'replace' of undefined.
Issue #1368 :Getting error: TypeError: Cannot read property 'replace' of undefined


reviewher:
f07a3d5fc0d7d7002d007a0753dc5c8f3f9d9798

Issue #1364, opened 2018-11-24
sheet_add_json with origin: -1 leaves gap from bottom of worksheet


dandv:
I'm seeing strange behavior with the following code:

```js
const XLSX = require('xlsx');

const wb = XLSX.utils.book_new();

const ws = XLSX.utils.json_to_sheet([
  { Product: 'p2', Qty: 2 },
]);

XLSX.utils.sheet_add_json(ws,
  [
    { Product: 'p3', Qty: 3 },
    { Product: 'p4', Qty: 4 },
    { Product: 'p5', Qty: 5 },
    { Product: 'p6', Qty: 6 },
    { Product: 'p7', Qty: 7 },
    { Product: 'p8', Qty: 8 },
  ],
  {
    header: ['Product', 'Qty'],
    skipHeader: true,
    origin: -1,  // append to bottom of worksheet starting on first column
  }
);


XLSX.utils.book_append_sheet(wb, ws, 'Bug sheet');

XLSX.writeFile(wb, 'bug.ods');
```

I expected `p3`...`p8` to be placed on rows 3 ... 8, but they ended up on rows 6 ... 11 (in LibreCalc at least).

![image](https://user-images.githubusercontent.com/33569/48968235-6353a780-efa1-11e8-98f0-3b5ed52b8eea.png)

I'm on 0.14.1. What might be going on?

dandv:
Did some more digging. Looks like if the JSON array has N >= 4 elements, the library inserts N - 3 empty rows, if the existing worksheet has 2 rows.

SheetJSDev:
The logical error is in https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L193-L199 : when `origin: -1` is specified and the number of rows inserted exceeds the number of rows in the file, it is added twice.

It would probably be simpler to proactively wraparound by checking if `_R < 0` in https://github.com/SheetJS/js-xlsx/blob/master/bits/90_utils.js#L186 and just adding the number of rows in the worksheet.  This would also handle cases like -2 for the penultimate row.

dandv:
Thanks for confirming and looking forward to the fix!

dandv:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=66589915)](https://www.bountysource.com/issues/66589915-sheet_add_json-with-origin-1-leaves-gap-from-bottom-of-worksheet?utm_medium=shield&utm_campaign=ISSUE_BADGE)

[...]

PR #1463, opened 2019-03-21
Fix for appending json to a sheet - #1364

dongido001:


SheetJSDev:
As noted in https://github.com/SheetJS/sheetjs/issues/1364#issuecomment-917878763 , the issue was fixed in a different way

PR #1516, opened 2019-05-24
perf package.json main field

reuwi:
To avoid ambiguity  and keep concise.


SheetJSDev:
Is there some environment where the current method of specifying the main script is causing problems?

reviewher:
9dda70c773c5f46aca24579fc03c7b513db66b0b

PR #1656, opened 2019-10-11
Fix for Issue: #1421 - TSV with empty first cell recognized wrong

### LINKED ISSUE #1421: TSV with empty first cell recognized wrong
When trying to parse a tab-separated-value sheet, where the first cell is empty - that means that the first byte is a '\t' character - the parser jumps into the wrong file format here https://github.com/SheetJS/js-xlsx/blob/1eb1ec985a640b71c5b5bbe006e240f45cf239ab/xlsx.js#L20310

jordonr:
Updated bits/87_read.js per SheeJSDev's comments (https://github.com/SheetJS/js-xlsx/issues/1421#issuecomment-537808818)

Made sure that I was able to perform the make process:
```console
$ git add xlsx.js
$ make clean
$ make
$ git diff xlsx.js
```

Closes #1421

### DIFF
--- a/bits/87_read.js
+++ b/bits/87_read.js
@@ -82,7 +82,7 @@ function readSync(data/*:RawData*/, opts/*:?ParseOpts*/)/*:Workbook*/ {
 	}
 	switch((n = firstbyte(d, o))[0]) {
 		case 0xD0: return read_cfb(CFB.read(d, o), o);
-		case 0x09: return parse_xlscfb(d, o);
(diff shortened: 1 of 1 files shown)


SheetJSDev:
You also have to remove the existing line.  In a diff view like

```diff
-		case 0x09: return parse_xlscfb(d, o);
+		case 0x09: if(n[1] <= 0x04) return parse_xlscfb(d, o); break;
```

Lines starting with a "-" appear in red and indicate lines that should be removed.  "+" lines appear in green and should be added.

I went ahead and made the change to the PR.

PR #1684, opened 2019-11-20
Fix for Issue: #1621 - Reading Excel HTML files will misalign cells in some cases

### LINKED ISSUE #1621: Reading Excel HTML files will misalign cells in some cases
I noticed a bug in the _html_to_sheet_ function. When reading cells that are empty, the cell index is correctly incremented, but if the cell has HTML tags and is observed to be empty after stripping the tags, the cell index is not incremented. This results in misalignment of the subsequent cells:

https://github.com/SheetJS/js-xlsx/blob/e3c5eac99c3b2be6929adfca455c1be87fab792b/xlsx.js#L18956-L18960

KurtMar:
Updated bits/79_html.js per SheetJSDev's comments (https://github.com/SheetJS/sheetjs/issues/1621#issuecomment-531007036), added a test case and verified with the SheetJS test files that no other tests failed.

Closes #1621

### DIFF
--- a/bits/79_html.js
+++ b/bits/79_html.js
@@ -36,9 +36,9 @@ var HTML_ = (function() {
 				/* TODO: generate stub cells */
 				if(!m.length) { C += CS; continue; }
 				m = htmldecode(m);
+				if(!m.length) { C += CS; continue; }
 				if(range.s.r > R) range.s.r = R; if(range.e.r < R) range.e.r = R;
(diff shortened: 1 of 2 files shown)


KurtMar:
@SheetJSDev Do I need to worry about this or is my work here done?

KurtMar:
@SheetJSDev is there something more needed for this pull request?

SheetJSDev:
b3793e2ea7ee6a6726e619373b1c287819fe7e2e

PR #1689, opened 2019-11-26
insert emptry string instead of skipping property

2aviv20:
sheet_to_json_skips_empty_rows_issue_139
https://github.com/SheetJS/sheetjs/issues/139


SheetJSDev:
This is intentional.  If a default value is not specified, the field should not be generated.  The proposed behavior can be implemented by setting `defval: ""`

PR #1703, opened 2019-12-12
fix: zero number format bug

suntopo:
## test demo in nodejs
```
var xlsx = require('node-xlsx');
var fs = require('fs');

const data = [[1, {t: 'n', f: 2, z: 2}, {t: 'n', f: 0, z: 2}]]

var buffer = xlsx.build([{name: "mySheetName", data: data}]); // Returns a buffer
fs.writeFileSync('test.xlsx', buffer)
```
## before change
2019-12-12 17-52-03屏幕截图
## after change
2019-12-12 17-53-03屏幕截图


SheetJSDev:
The issue here is that `f` should be a string, not a number:

```js
const data = [[1, {t: 'n', f: "2", z: 2}, {t: 'n', f: "0", z: 2}]]
```

So it's preferable to check if it is a string

Issue #1717, opened 2020-01-06
Named cell reference causes the string "undefined" to be output in the ODS file


dandv:
```js
const XLSX = require('xlsx');

const wb = XLSX.utils.book_new();

const ws = XLSX.utils.json_to_sheet([]);

XLSX.utils.book_append_sheet(wb, ws, 'Bug sheet');

ws.A1 = { t: 's', f: 'square' };

XLSX.writeFile(wb, 'formula.ods');
```

The resulting .ODS file contains:

    <table:table-cell table:formula="of:=square" office:value-type="string"><text:p>undefined</text:p></table:table-cell>

Not sure what `<text:p>` is (probably a placeholder?), but it should simply be omitted in this case.

Adding a `v: ''` to the cell object outputs an empty `<text:p></text:p>`.

    <table:table-cell table:formula="of:=square" office:value-type="string"><text:p></text:p></table:table-cell>

That, again, should be omitted, or optimized to `<text:p/>`.

SheetJSDev:
This is actually an issue with a string cell that has no value.  The fix is pretty straightforward (feel free to submit a PR): https://github.com/SheetJS/sheetjs/blob/master/bits/81_writeods.js#L81

```diff
-						textp = cell.v;
+						textp = cell.v == null ? "" : cell.v;
```

PR #1727, opened 2020-01-23
fix: #1717

djrumph:
I saw the issue was raised and a SheetJS Dev said the fix was simple and involved dealing with defined and undefined variables.
They wrote a line using a ternary conditional to assign the variable textp to the value of cell.v if it's defined, or an empty string if it is not defined. 

SheetJS Dev Code: textp = cell.v == null ? "" : cell.v;

Suggesting using the line: textp = cell.v || "";

This does the same thing and assigns textp to cell.v if it is defined or an empty string if it is not. 

(Note any feedback would be greatly appreciated, this is my first try at contributing to open source).

### DIFF
--- a/bits/81_writeods.js
+++ b/bits/81_writeods.js
@@ -78,7 +78,7 @@ var write_content_ods/*:{(wb:any, opts:any):string}*/ = (function() {
 						ct['office:value'] = (cell.v||0);
 						break;
 					case 's': case 'str':
-						textp = cell.v;
+						textp = cell.v || "";
 						ct['office:value-type'] = "string";
 						break;
 					case 'd':


SheetJSDev:
In JS there are 5 values that evaluate to false in a boolean context: `undefined`, `null`, `0`, `false` and the empty string.  The interesting (and as of yet unanswered) question is how to handle string cells with a numeric value of `0` or a boolean value of `false`.  Your code would render both values as empty strings, whereas my code would render the numeric `0` as `"0"` and the boolean `false` as `"false"`.  Thoughts?

djrumph:
Thank you for the great feedback and opportunity to dig into the code. I really appreciate your time.  
So the goal is to only render an empty string if the value of cell.v is undefined, null or NaN, otherwise convert the value of the string to a primitive and assign it to textp?

If so here is my idea, feel free to skip over this if I interpreted what you meant wrong.  

IDEA:
My thoughts are to use the original ternary statement combined with extra checks and a function to convert the cell.v value to a primitive. 

Ultimately the code that passed all tests was:

textp = (cell.v === undefined || cell.v  === null || isNan(cell.v.toString)) ? "" : JSON.parse(cell.v);

This code assigns the primitive value of the cell to the variable unless it is undefined, null or NaN.
I added a isNaN(cell.v.toString) to pass both the NaN test together. IsNaN() without toString() failed the 0n test. 

The function JSON.parse turns a string into JSON, and for this purpose, it essentially turns strings into primitives. 


END
Let me know if this addresses the issue you brought up or if I missed the mark.

Thank you again for the feedback. I really enjoy learning more by solving problems.

reviewher:
57b296e5b5fff4300de1538fb802114cceb67991

Issue #1779, opened 2020-02-28
XLSX.stream.to_json doesn't ends when a blank row is the last row in Excel sheet.


rohandhamapurkar:
### sheetjs/bits/97_node.js
```
		stream._read = function() {
 			if(R > r.e.r) return stream.push(null);
			while(R <= r.e.r) {
				//if ((rowinfo[R-1]||{}).hidden) continue;
				var row = make_json_row(sheet, r, R, cols, header, hdr, dense, o);
				++R;
				if((row.isempty === false) || (header === 1 ? o.blankrows !== false : !!o.blankrows)) {
					stream.push(row.row);
					break;
				}
			}
		};
		return stream;
```

Hi @SheetJSDev !

This piece of code has the ```R > r.e.r``` checking outside of the loop but when a blank row is inserted at the end of a given sheet then the ```if((row.isempty === false) || (header === 1 ? o.blankrows !== false : !!o.blankrows)) ``` condition inside of the while loop isn't executed and the stream does not end.

Ill be submitting a pull request for the same for a fix for this issue that I have found.
Please verify and merge the changes as soon as possible as I am using this library in production in my system.

highflying:
I'm seeing the same issue, for now I'm having to specify the range to skip processing the blank lines at the end.

rohandhamapurkar:
@highflying can you please elaborate more when you say 
"specify the range to skip processing the blank lines"

highflying:
> @highflying can you please elaborate more when you say
> "specify the range to skip processing the blank lines"

I know the range of the sheet that has data in it, so I am specifying it manually, e.g.

```
  const stream = XLSX.stream.to_json(sheet, {
    range: "A1:BN3",
  });
```

When doing this the `end` event gets emitted, without it no `end` event happens.

rohandhamapurkar:
Oh makes sense thanks a lot.
Please do make this change live soon I use this in production.
Thanks again for your response.

ev45ive:
Ok. If i have many files with variable number of rows and I want to stream it as those are big files.

How do I end the stream then?

rohandhamapurkar:
> Ok. If i have many files with variable number of rows and I want to stream it as those are big files.
> 
> How do I end the stream then?
@ev45ive 
Every worksheet object has "!range" property which can be passed in to the ```XLSX.stream.to_json``` function and you can do it dynamically.

ev45ive:
WIth range option stream still seams not to end
```
[...]

PR #1782, opened 2020-02-29
fix: #1779

rohandhamapurkar:
Probably fixes XLSX.stream.to_json doesn't ends when a blank row is the last row in Excel sheet
@SheetJSDev please review and merge the changes!


SheetJSDev:
467020fc693f1c34cfa32670022c3f8ef4199a58

Issue #1883, opened 2020-03-26
refactor this code to use modern code (es6+) and typescript?


sibelius:
are you willing to accept pull requests to migrate incrementally to typescript or more modern code like es6+?

I find kind of "weird" the bits folder (is there all the code is located right now?)

can we also move it inside `packages\xlsx`?

SheetJSDev:
When the question was posed years ago, the tooling wasn't great.  Older versions of VSCode and github just choke on the `xlsx.flow.js` script among others.   You can see it in action, https://github.com/SheetJS/sheetjs/blob/master/xlsx.flow.js takes seconds to load and the blamelog literally gives you a 504

<img width="481" alt="blamelog 504" src="https://user-images.githubusercontent.com/6070939/77706518-fab37b00-6f98-11ea-888b-d483d76586b6.png">

The tooling has improved a lot since then, and I think we could start "modernizing" the codebase.  The key requirements are preserving IE compatibility in a standalone script and reproducible builds (easy to do when you're just concatenating scripts)

For historical reasons, the project is split up into multiple repos (The original discussion was in the xls repo and transferred to https://github.com/SheetJS/sheetjs/issues/1834).  Now that the two parts are merged in one project, we should move all of the parts into the repo, and I agree that this should live in the packages directory.

So to start, we could probably make a TS version of `bits/04_base64.js`

arantespp:
I'd like to help with this refactor. What would be the first step? Starting with the TS version of `bits/04_base64.js`, the idea is to create a `bits/04_base64.ts`?

SheetJSDev:
An ESM build has been added to the repo.  It's not 100% yet due to limitations of ES6 import.

PR #1900, opened 2020-04-13
fix: #1883 Add TypeScript

arantespp:
Tasks:

- create a TypeScript first configuration
- create `tsbuild` npm script
- create the file `bits/04_base64.ts`

With this PR, I proposed to create the file `bits/04_base64.ts` and the corresponding `.js` is generated running the command `npm run tsbuild`, which will create `bits/04_base64.js` file.

I've kept the original `tsconfig.json` file to be easier to adjust if we want to. When the final configuration arrives, we delete all commented properties.


arantespp:
#1883

reviewher:
3cbe83c8559b521b6cb19e53de64a11eac6c36a0

PR #1908, opened 2020-04-20
fix: missing PRN in xlsx.mini

chenxeed:
# Issue

https://github.com/SheetJS/sheetjs/issues/1760

# Description

Based on my investigation, it seems like the function `read_prn` in the mini is missing `PRN` when it tries to read a string of CSV file.

My proposed solution is to add the `PRN` declaration as part of the `mini` build, which exists in the `bits/40_harb.js`.

### DIFF
--- a/mini.lst
+++ b/mini.lst
@@ -22,6 +22,7 @@ bits/31_rels.js
 bits/33_coreprops.js
 bits/34_extprops.js
 bits/35_custprops.js
+bits/40_harb.js
 bits/42_sstxml.js
 bits/46_stycommon.js
 bits/47_styxml.js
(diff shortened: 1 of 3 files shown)


chenxeed:
@SheetJSDev I have no written unit test in this changes, but I have tested the changes in my local application and the CSV works.

test import svg

chenxeed:
Poke @SheetJSDev again 🙇 it would be great to have this settled for the next release, so we can use `xlsx.mini` in our application ☕

chenxeed:
Hi @SheetJSDev @srijonsaha may I have an update or response for the PR?

If there's anything missing from the PR, please let me know to improve.

I'm waiting for this patch before I can use `xlsx.mini` in my production project.

Thanks!

SheetJSDev:
@garrettluu can you take a peek?

chenxeed:
Thanks @garrettluu for the response and review. I've squashed the commit. Look forward to the update 👍

chenxeed:
Hi @garrettluu @SheetJSDev I've made the requested changes. Please confirm if this is fine. Look forward to the release! 🙇

PR #1938, opened 2020-05-10
feat: add writeFileAsync definition to typescript

whlsxl:


SheetJSDev:
bb9faf7caf9e163f0a96767eaa9c8c1a8a8bc350

tslint was not happy with the bare `Function` type so it was replaced with `()=>void`

PR #2009, opened 2020-06-17
Add option to force quotes around values in when exporting to csv

### LINKED ISSUE #1694: Force quotes around strings when exporting CSV
In the default setting, quotes are only added around strings if they’d collide with the separator character (i.e. string contains comma). Is there a configuration option which would allow me to always put quotes around strings?

enghong:
Fix #1694

### DIFF
--- a/README.md
+++ b/README.md
@@ -2140,14 +2140,15 @@ For the example sheet:
 As an alternative to the `writeFile` CSV type, `XLSX.utils.sheet_to_csv` also
 produces CSV output.  The function takes an options argument:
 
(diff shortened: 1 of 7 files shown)


SheetJSDev:
Looks good, but is there a specific reason you need to quote every single field? Is there some corner case where SheetJS isn't quoting a value when it should?

enghong:
Not at all, SheetJS is quoting when it should.

I just have a client who explicitly need all values to be quoted, I'm not sure why either...
Then I found the #1694, so I decided to implement the option.

PR #2014, opened 2020-06-18
adding an option to not parse numbers

ab320012:
i have a problem that this PR will solve let me know if it needs anything else to be included.

### DIFF
--- a/bits/90_utils.js
+++ b/bits/90_utils.js
@@ -33,7 +33,7 @@ function make_json_row(sheet/*:Worksheet*/, r/*:Range*/, R/*:number*/, cols/*:Ar
 				else if(raw && v === null) row[hdr[C]] = null;
 				else continue;
 			} else {
-				row[hdr[C]] = raw ? v : format_cell(val,v,o);
+				row[hdr[C]] = raw || (o.rawNumbers && val.t == "n") ? v : format_cell(val,v,o);
 			}
 			if(v != null) isempty = false;
 		}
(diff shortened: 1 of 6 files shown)


SheetJSDev:
Let's take a step back.

1) What you changed affects the CSV writer, not the CSV parser.  Is that the intention?

2) Is the goal specifically to write raw numbers or should it also write raw strings / boolean / dates as well?

3) If the goal is to change the CSV writing, the changes should only be in `make_csv_row`.  I would have written something like:

```diff
-			txt = ''+format_cell(val, null, o);
+			txt = ''+(o.rawNumbers && val.t == "n" ? val.v : format_cell(val, null, o));
```

ab320012:
i just need to be able to write the numbers raw but keep the formatting the same otherwise i cna make the requested change

ab320012:
i'll add it to sheet_to_json as well just in case?

SheetJSDev:
`sheet_to_json` already has the option `raw`.  When true (default) it will give the underlying values, and when false (have to explicitly pass the option) it will use the formatted text.  The difference between `raw` and your proposed `rawNumbers` is that the former works for all types and the latter is only for numbers (hence question 2)

ab320012:
Hey yes it's just to write raw numbers but keep the other formatting, I will make the change you asked

ab320012:
@srijonsaha great library, not having this is causing some production issues with our website wondering if you can approve

ab320012:
@SheetJSDev @srijonsaha hey just bumping this i still need it for my prod application

srijonsaha:
@ab320012 Hi could you give a common example use case of what this change would be? We don't want to accept every change simply because it solves a very specific one time situation.

ab320012:
@srijonsaha yes absolutely, the number parser right now is formatting my csv in scientific notation, i'd like to keep the numbers raw but parse everything else the same way that it was, is that enough of a use case? it's just an added option so i can use the internal parser but keep the numbers the way they are does that make sense?

ab320012:
@srijonsaha hey thanks for the merge this is causing us some prod issues, wondering if you will publish soon?

PR #2021, opened 2020-06-22
Firebase Functions demo

garrettluu:
Added a demo that runs on Firebase functions to convert Excel files to `.csv`. Takes a file from an HTTP request as input and outputs the csv text as a response.


garrettluu:
Will be combined with #2025

PR #2029, opened 2020-06-27
xlsx-cli refactor

### LINKED ISSUE #1922: [RFC] Split out the CLI into a separate package
Currently, some of the dependencies of this package are related to its CLI tool (which is not even mentioned in the readme AFAICT). These dependencies (`commander` and `exit-on-epipe` at least) are unnecessary for projects using the library. To reduce the install footprint, what do you think about splitting the CLI into a separate package (that can then be installed only when you actually need it) ?

Note that to actually benefit from the smaller install footprint, this should also be done in other packages you maintain (`crc-32`, `adler-32`, `cfb` and `codepage`), as they also include their own CLI with these dependencies (btw, currently, this forces the install of 2 different versions of `commander` due to `xlsx` using an older version than `codepage`)

garrettluu:
Closes #1922 

Refactored the CLI into `packages/xlsx-cli` and published the NPM package [here](http://npmjs.com/package/xlsx-cli).
With this refactor, we have completely `commander` and `exit-on-epipe` as dependencies in the main library and the submodules.

For convenience, here are the PRs for the CLI refactors of the submodules in one place:
- [alder32](https://github.com/SheetJS/js-adler32/pull/4)
- [ssf](https://github.com/SheetJS/ssf/pull/51)
- [crc32](https://github.com/SheetJS/js-crc32/pull/15)
- [cfb](https://github.com/SheetJS/js-cfb/pull/10)

To run the cli, install it with `npm i -g xlsx-cli`, then run `npx xlsx-cli`.


packages/xlsx-cli/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "bin": {
+    "xlsx-cli": "./bin/xlsx.njs"
+  },
+  "dependencies": {
+    "commander": "^5.1.0",
+    "exit-on-epipe": "^1.0.1",
+    "fs": "^0.0.1-security",
+    "xlsx": "^0.16.2"
+  },
+  "devDependencies": {},
```
srijonsaha, line 15: this line could be removed (minor thing)

stof:
The xlsx-cli package.json should have the `repository` config, so that npmjs.com can link to the source repository.

stof:
@garrettluu the `codepapage` package also needs such a refactoring

reviewher:
c1f5f04

PR #2042, opened 2020-07-03
fix: change test to read string, not buffer

wlawt:
The browser test `CSV > input > should interpret CRLF newlines` was throwing an error for buffer not being defined.

A quick fix was to read it as a `string` instead of a `buffer`.

This also passes the browser tests and the Sauce Labs tests.

### DIFF
--- a/tests/core.js
+++ b/tests/core.js
@@ -1973,7 +1973,7 @@ describe('CSV', function() {
 			assert.equal(get_cell(sheet, "C1").v, '100');
 		});
 		it('should interpret CRLF newlines', function() {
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Relatively recently.  When we originally made the test for nodejs, Buffer was used.  That's not available in browser.  @wlawt found that the some of the node tests weren't be run in-browser, and tried reconciling the two test suites in a previous commit.  He tried to run the browser tests and found this issue, and the PR addresses the problem (https://app.saucelabs.com/builds/c1670f5dd3384c968095e8ed797c3d57 is the browser test run)

PR #2053, opened 2020-07-08
Update 20_import.md [ci skip]

jrmdel:
When creating the workbook, need to retrieve result from `reader` instead of the event (`e.target`)

See : https://webplatform.github.io/docs/apis/file/FileReader/readAsArrayBuffer/#Examples

### DIFF
--- a/docbits/20_import.md
+++ b/docbits/20_import.md
@@ -109,7 +109,7 @@ function handleDrop(e) {
   var files = e.dataTransfer.files, f = files[0];
   var reader = new FileReader();
   reader.onload = function(e) {
-    var data = new Uint8Array(e.target.result);
+    var data = new Uint8Array(reader.result);
     var workbook = XLSX.read(data, {type: 'array'});
 
     /* DO SOMETHING WITH workbook HERE */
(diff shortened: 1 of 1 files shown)


SheetJSDev:
The actual specification https://www.w3.org/TR/FileAPI/#intro uses event target rather than the out-of-scope reader variable.  From the intro:

```js
function loaded(evt) {
  // Obtain the read file data
  var fileString = evt.target.result; // <-- event target
  // Handle UTF-16 file dump
  if(utils.regexp.isChinese(fileString)) {
    //Chinese Characters + Name validation
  }
  else {
    // run other charset test
  }
  // xhr.send(fileString)
}

function getAsText(readFile) {

  var reader = new FileReader();

  // Read file into memory as UTF-16
  reader.readAsText(readFile, "UTF-16");

  // Handle progress, success, and errors
  reader.onprogress = updateProgress;
  reader.onload = loaded;
  reader.onerror = errorHandler;
}
```

[Section 6.4 of the spec](https://www.w3.org/TR/FileAPI/#events) assures that this works:

> The FileReader object must be the event target for all events in this specification.

PR #2069, opened 2020-07-23
Fix rawNumber support inside sheet_to_json

### LINKED ISSUE #2068: rawNumbers: false does not work in sheets_to_json in combination with raw: true
Use case: parse dates appropriately with `cellDates: true` but show numbers with formatted number strings `raw:true, rawNumbers: false`

delesseps:
Resolves #2068 and depends on a test_file being added via: [test_file PR](https://github.com/SheetJS/test_files/pull/11)


delesseps:
@SheetJSDev   I believe this fixes a problem that was recently added in #2014.  cc @ab320012   Please let me know if it seems ok.  As mentioned I added a related test file that needs to be merged for the tests to pass.

SheetJSDev:
Let's start from scratch.  @delesseps suppose none of these options existed.  What would be the appropriate options and defaults?  Here are the main considerations:

1) Dates may be represented as date objects or as Excel Date Codes with appropriate number format, so there probably should be an option to control how dates and date-like numbers are processed

2) Excel separates "content" (raw value) from "presentation" (displayed text), so there probably should be an option to control whether the function returns the raw values or the formatted text

Questions:

- Should the content/presentation issue be adjustable at the cell type level (number, date, string, boolean, error) or should it be a single option (like `raw` currently)?

- What is the appropriate default behavior when a worksheet has both true date cells (type `d`) and date-like numbers (type `n` with a date number format)?  Should it try to normalize as dates, normalize as numbers, or leave as-is?

geektcp (commented):
LGTM

JSilversun:
Hello, I'm parsing big excel files and I would like to get the numbers as a raw string to prevent conversions to javascript `Number` object to prevent any kind of precision loss, is it possible to parse an excel file to JSON and read the values as raw strings but with all decimals? Thank you

SheetJSDev:
Thank you very much!

https://github.com/SheetJS/sheetjs/commit/a5b387716c2c3ec9b9794c5712a80f857e52a5d3 main repo commit

https://github.com/SheetJS/test_files/commit/2571413d2560c1a41a73d9515e20aaac99c34570 test files commit

PR #2072, opened 2020-07-27
Fix encoding wb.WBProps.CodeName

iroot:
Without this fix does not work macros in Excel in the national language.

### DIFF
--- a/bits/72_wbxml.js
+++ b/bits/72_wbxml.js
@@ -37,7 +37,7 @@ function parse_wb_xml(data, opts)/*:WorkbookFile*/ {
 						default: wb.WBProps[w[0]] = y[w[0]];
 					}
 				});
-				if(y.codeName) wb.WBProps.CodeName = y.codeName;
+				if(y.codeName) wb.WBProps.CodeName = utf8read(y.codeName);
 				break;
 			case '</workbookPr>': break;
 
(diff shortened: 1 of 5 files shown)


SheetJSDev:
This looks good.  Can you give an example of a real non-ASCII code name so we can use in a test?  For example, using `本工作簿`, the following test fails against HEAD but passes with your changes:

```js
var wb = XLSX.utils.book_new();
var ws = XLSX.utils.aoa_to_sheet([[1]]);
XLSX.utils.book_append_sheet(wb, ws, "Sheet1");
wb.Workbook = { WBProps: { CodeName: "本工作簿" } };
var bstr = XLSX.write(wb, {type: "binary", bookType:"xlsx"});
var nwb = XLSX.read(bstr, {type: "binary"});
assert.equal(nwb.Workbook.WBProps.CodeName, wb.Workbook.WBProps.CodeName);
```

iroot:
In my case, the Cyrillic alphabet is used and the CodeName will be equal to `ЭтаКнига`.
```js
var wb = XLSX.utils.book_new();
var ws = XLSX.utils.aoa_to_sheet([[1]]);
XLSX.utils.book_append_sheet(wb, ws, "Sheet1");
wb.Workbook = { WBProps: { CodeName: "ЭтаКнига" } };
var bstr = XLSX.write(wb, {type: "binary", bookType:"xlsx"});
var nwb = XLSX.read(bstr, {type: "binary"});
assert.equal(nwb.Workbook.WBProps.CodeName, wb.Workbook.WBProps.CodeName);
```

PR #2146, opened 2020-10-14
Issue#2139 | fix for json_to_sheet - mutates header array

tomRzeznik:


tomRzeznik:
Fix for header array mutation

himanshusaini111:
No Fix is needed the behaviour was intentional, as mentioned in the issue discussion.

SheetJSDev:
The behavior is intentional.  https://github.com/SheetJS/sheetjs/issues/2139#issuecomment-706763053 explains it further.  The "fix" is to clarify the documentation.  As the discussion revealed, many libraries behave differently and a line in the documentation would help set expectations.

PR #2166, opened 2020-11-02
Add support for configuring outlines to be above the detail row, or to left of detail column

robinfhu:
This adds support for the following Excel feature (in the Group configuration):

Screen Shot 2020-11-02 at 3 20 57 PMEST

Allowing the following output:

Screen Shot 2020-11-02 at 4 55 03 PMEST


AshtomGVA:
Thank you for this!
Can we know when it is going to be released?

PR #2184, opened 2020-11-24
Fix Types not Resolving in Node ESModule projects

josh-hemphill:
EsModule imports do not automatically resolve index files of directories, so the `"types": "types"` does not resolve to the `types/index.d.ts` as was intended. If it doesn't break anything, it should be changed to `"types": "types/index.d.ts"`.
According to the [Typescript Spec](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html), the types/typings field should be a file, not a directory.


SheetJSDev:
Does it actually require the filename?  If so, how was this even working in the first place?

josh-hemphill:
Sorry, I though the problem might just be my configuration, but It's gone full circle now, and nothing I've been able to do has gotten me intellisense/type completion.
If I set `baseUrl` then VSCode gives me `File "[Absolute Path]/node_modules/xlsx/types" not found` in my jsconfig file. 
If I do `/// <reference path="../../../node_modules/xlsx/types/index.d.ts" />` in the file I need it, I do get the type information.
The reason this probably only came up now is the native ES Module support in Node; with the new ES Module support, if you make your project an ES Module project, then there won't be any `/index.*` inference since ES Imports require absolute paths when not being compiled by typescript.

nyenye:
Will there be  a new release with this patch? I find myself with the same issue, and even thought we can fix it after `npm install` it would be great if it solution could be shared among the team, with no need of hacking it away. Thanks a lot for the solution :)

SheetJSDev:
Reasoning makes sense, and there's no evidence to suggest any tooling expects `types` or `typings` to be a folder.

326b7c0ca18970afe9494eda78488d20888c4453

PR #2291, opened 2021-05-13
Add option to force quotes around string values when exporting to csv

enghong:
This is very similar to [486f35b](https://github.com/SheetJS/sheetjs/commit/486f35b4cc59d53d5c046760fe98923ade9201ad) but quotes will be put only around string fields (and not numbers).

I have a client who explicitly requires this, not sure why...

I'm submitting this in case it can be useful to others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2201,15 +2201,16 @@ For the example sheet:
 As an alternative to the `writeFile` CSV type, `XLSX.utils.sheet_to_csv` also
 produces CSV output.  The function takes an options argument:
 
-| Option Name  |  Default | Description                                        |
-| :----------- | :------: | :------------------------------------------------- |
-|`FS`          |  `","`   | "Field Separator"  delimiter between fields        |
-|`RS`          |  `"\n"`  | "Record Separator" delimiter between rows          |
-|`dateNF`      |  FMT 14  | Use specified date format in string output         |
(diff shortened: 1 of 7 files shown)


SheetJSDev:
This is probably useful for your use case, but a few issues need to be discussed (writing code is the easy part):

- The option is `forceQuotesAroundString` but the actual behavior is forcing quotes around everything that is not a number.  For example, logical true/false would be wrapped in quotes.  Is that desirable behavior?

- Should this option be rolled into `forceQuotes`?  If the value is not true/false or 0/1, the value can be interpreted as a bitfield where different bits correspond to different field types (e.g. `2` for numeric cells, `4` for string cells, `8` for boolean cells, etc).  Under this system, a value like `12 = (4 | 8)` would mean that string and boolean cells would have forced quotes but not the other cell types.

- if `forceQuotesAroundString` is better kept as a separate option, what should the combination `forceQuotes: true, forceQuotesAroundString: false` do?

enghong:
- In the CSV context, I think it makes sense to only consider two types: `number` and `string`, at least it was what I though.
- If we agree on the previous point, we can modify a bit `forceQuotes` option if the value is not true/false, it can be interpreted as string:
  - `'all'` which is equivalent to `forceQuote: true`
  - `'string'` for all non-numeric cells

Would that clarify the usage?

PR #2296, opened 2021-05-22
avoid filereader

### LINKED ISSUE #2295: Discourage use of the FileReader
First and formost i would like to discourage the use of `readAsBinaryString`

Javascript dose not handle binary strings particular well...

> Note that this method was once removed from the File API specification, but re-introduced for backward compatibility.
Using FileReader.readAsArrayBuffer() is recommended.<sup>[[2][2]]</sup>

- If they think it's okay to use a binary string then they might think that getting a binary string from XHR or `fetch(url).then(res => res.text())` works well also - when in fact it dose not, ajax performs text encoding on binary strings when being fetched<sup>[[6][6]]</sup>
- They removed `readAsBinaryString` from the spec for a reason, so we should stop using it.
- When the ArrayBuffer interface appeared, readAsBinaryString has been deprecated because all its use cases could be done in a better way using that new interface.
Indeed, readAsBinaryString only converts the binary data into a DOMString (UTF-16). There is not much you can do from it afterward. Also, storing it as an UTF-16 string implies that it takes far more space in memory than the original data size. Add to this that strings are immutable, I guess you can see how inefficient it is to work from this. <sup>[[1][1]]</sup>
- Also, chrome can't handle 500+ MiB strings but buffers can<sup>[[3][3]]</sup>

I do not know what the reason behind using a binary string is, but if you for some reason need to get some part of the chunk as a text then a better conversion would then be to use `str = new TextDecoder().decode(data.slice(0, 1024))`?
The sheets are also in binary format so maybe it would be easier to parse the binary using DataView. (Using the DataView instead of Buffer would reduce the dependency on the browser build)

---

Now for the 2nd reason of why not to encourage the use of FileReader any more:

- NodeJS v15.7 introduced support for Blobs<sup>[[4][4]]</sup>. They are looking into adding support for getting blobs from the FileSystem as well.
- What NodeJS don't intend to implement is the old `FileReader`. It's a old legacy API now days when there is this new Blob reading methods supported directly onto the blob<sup>[[5][5]]</sup> 
- I think the new Blob read methods have a simpler syntax with the use of async/await + promises
- And everyone will be able to read any instance of any Blob polyfill like the [fetch-blob](https://github.com/node-fetch/fetch-blob) package for example
  - Not every FileReader Polyfill would be able to read any Blob Polyfill since they would probably read some internal/private `_buffer` property if they don't use the new read methods

If we instead encouraged everyone to use Uint8Array instead then i think it would have a better cross platform support on Deno, NodeJS, and also in the browser. (if it's even possible to use sheetjs on the server - i don't know)

It would also make the code simpler if there is just one way do read the data instead of multiple ways to reduce the code complexity
It would also make it easier for PPl to know how to import data from ajax since it has no way of getting data as binary string

---

I'm willing to help with PR's if we can change any occurrence of FileReader 
```diff
-      $elm.on('change', function (changeEvent) {
-        var reader = new FileReader();
-
-        reader.onload = function (e) {
-          /* read workbook */
-          var bstr = e.target.result;
-          var workbook = XLSX.read(bstr, {type:'binary'});
-
-          /* DO SOMETHING WITH workbook HERE */
-        };
-
-      reader.readAsBinaryString(changeEvent.target.files[0]);
-     })

into something like this:

+      $elm.on('change', async evt => {
+        const file = evt.target.files[0];
+        const buffer = await file.arrayBuffer();
+        const workbook = XLSX.read(new Uint8Array(buffer), { type: 'array' });
+       });
```

I could do multiple smaller PR's for each file if that is easier to review, or i could do several at once.

I could also help create some depreciation warning to the console if anyone are using the binary, 
`XLSX.read(data, { type: 'binary' })` syntax

I would also like to emphasize that the read method could be a tiny bit smarter if it could do auto detection based on the kind of data it is. instead of setting `{ type: 'array' }` it could instead switch based on the first argument and look if it is a ArrayBuffer or ArrayBufferView.

Then it could be a tiny bit easier to just do:
```js
$elm.on('change', async evt => {
  const file = evt.target.files[0]
  const workbook = XLSX.read(await file.arrayBuffer())
})

  const res = await fetch(url)
  const workbook = XLSX.read(await res.arrayBuffer())
```

 [1]: https://stackoverflow.com/questions/55836771/why-is-readasbinarystring-deprecated
 [2]: https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsBinaryString
 [3]: https://stackoverflow.com/questions/67483360/chrome-filereader-returns-empty-string-for-big-files-300mb/67501438#67501438
 [4]: https://nodejs.org/dist/latest-v16.x/docs/api/buffer.html#buffer_class_blob
 [5]: https://developer.mozilla.org/en-US/docs/Web/API/Blob#instance_methods
 [6]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Sending_and_Receiving_Binary_Data#receiving_binary_data_in_older_browsers

jimmywarting:
Fixes #2295

### DIFF
--- a/README.md
+++ b/README.md
@@ -415,20 +415,13 @@ req.send();
 <details>
   <summary><b>Browser drag-and-drop</b> (click to show)</summary>
 
-Drag-and-drop uses the HTML5 `FileReader` API.
+Drag-and-drop to read blobs.
 
 ```js
-function handleDrop(e) {
-  e.stopPropagation(); e.preventDefault();
-  var files = e.dataTransfer.files, f = files[0];
-  var reader = new FileReader();
-  reader.onload = function(e) {
-    var data = new Uint8Array(e.target.result);
-    var workbook = XLSX.read(data, {type: 'array'});
-
-    /* DO SOMETHING WITH workbook HERE */
-  };
-  reader.readAsArrayBuffer(f);
+async function handleDrop(evt) {
+  evt.stopPropagation(); evt.preventDefault();
+  const file = evt.dataTransfer.files[0];
+  const workbook = XLSX.read(await file.arrayBuffer());
 }
 drop_dom_element.addEventListener('drop', handleDrop, false);
 ```
@@ -438,20 +431,13 @@ drop_dom_element.addEventListener('drop', handleDrop, false);
 <details>
   <summary><b>Browser file upload form element</b> (click to show)</summary>
 
-Data from file input elements can be processed using the same `FileReader` API
+Data from file input elements can be processed the same way
 as in the drag-and-drop example:
 
 ```js
-function handleFile(e) {
-  var files = e.target.files, f = files[0];
-  var reader = new FileReader();
-  reader.onload = function(e) {
-    var data = new Uint8Array(e.target.result);
-    var workbook = XLSX.read(data, {type: 'array'});
-
-    /* DO SOMETHING WITH workbook HERE */
-  };
-  reader.readAsArrayBuffer(f);
+async function handleFile(evt) {
+  const file = evt.target.files[0];
+  const workbook = XLSX.read(await file.arrayBuffer());
 }
 input_dom_element.addEventListener('change', handleFile, false);
 ```
--- a/demos/angular/README.md
+++ b/demos/angular/README.md
@@ -81,19 +81,10 @@ app.directive("importSheetJs", [SheetJSImportDirective]);
 function SheetJSImportDirective() {
   return {
     scope: { opts: '=' },
-    link: function ($scope, $elm) {
-      $elm.on('change', function (changeEvent) {
-        var reader = new FileReader();
-
-        reader.onload = function (e) {
-          /* read workbook */
-          var bstr = e.target.result;
-          var workbook = XLSX.read(bstr, {type:'binary'});
-
-          /* DO SOMETHING WITH workbook HERE */
-        };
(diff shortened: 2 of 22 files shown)


jimmywarting:
tough to know what the failing test are when you can't see the logs...

jimmywarting:
bump?

jimmywarting:
Someone wants to take a look at this?

SheetJSDev:
It'll be split up into multiple commits as we review the demos.  The first commit 36a1a4e9b82a605412df3c365cf64de86f16291c addresses the react demo and has been set to the author metadata from this PR.

SheetJSDev:
Slowly starting to remove `readAsBinaryString` from the demos.

[`Blob#arrayBuffer` might be too new for the demos](https://caniuse.com/mdn-api_blob_arraybuffer) and would require a Promise library for use in IE10+, so the best approach is probably to include a note about the new API in the main README

SheetJSDev:
`FileReader#readAsBinaryString` has been completely replaced with `FileReader#readAsArrayBuffer`.

Some of the native demos did not support `Blob#arrayBuffer`, so most of the demos still use FileReader.  The README now includes demos for both approaches.  When more of the ecosystem supports the new API, we will revisit changing the demos again.

Issue #2299, opened 2021-05-27
xlsxworker.js points to shim.js instead of dist/shim.min.js


btopro:
effectively trying to replicate the demo found here https://oss.sheetjs.com/sheetjs/ but with the latest 0.17.0 release of the library. In local usage if `importScripts('shim.js');` is instead moved to `importScripts('dist/shim.min.js');
` it works just as the demo out on the web does. PR shortly.

sibelius:
is there a changelog for 0.17.0?

SheetJSDev:
5a3e5d91c1bf1aff548deca292be2c95a304cb11

@sibelius https://github.com/SheetJS/sheetjs/issues/1728

PR #2300, opened 2021-05-27
fix for #2299

btopro:
Simple fix to point to the actual location for the webworker's shim file which is in the dist directory.


SheetJSDev:
5a3e5d91c1bf1aff548deca292be2c95a304cb11

PR #2303, opened 2021-05-29
Fix:Non-mini xlsx.js missing "Root Entry" removal for proper ZIP parsing

### LINKED ISSUE #1736: Non-mini xlsx.js missing "Root Entry" removal for proper ZIP parsing
There is a difference in `safegetzipfile` method between [xlsx.js](https://github.com/SheetJS/sheetjs/blob/a81bb78f18960053a03bb3350f31ea54d28245be/xlsx.js#L2810) & [xlsx.mini.js](https://github.com/SheetJS/sheetjs/blob/a81bb78f18960053a03bb3350f31ea54d28245be/xlsx.mini.js#L2789).

One is missing `replace(/^Root Entry[\/]/,"")` so I get this error "Cannot find file [Content_Types].xml in zip" every time trying to read a XLSX file 🤷‍♂️

Patil2099:
Fixes #1736

### DIFF
--- a/bits/21_ziputils.js
+++ b/bits/21_ziputils.js
@@ -29,7 +29,7 @@ function safegetzipfile(zip, file/*:string*/) {
 	var k = zip.FullPaths || keys(zip.files);
 	var f = file.toLowerCase(), g = f.replace(/\//g,'\\');
 	for(var i=0; i<k.length; ++i) {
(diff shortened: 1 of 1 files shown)


reviewher:
ef4aec8796c72dc891b619edcd3ef24dde4829d6

PR #2331, opened 2021-07-22
Use opts.codepage as default encoding for DBF

dnafigator:


SheetJSDev:
There were a few questions at the time for which we didn't have compelling answers:

1) How should unsupported codepages be handled?  DBF only supports a small subset of the codepages supported by XLS and other formats.  "Common" codepages like 1257 ("Windows Baltic" designed for Lithuanian and Latvian) are not supported.  Should an error be thrown or should it default to 1252?

2) Newer versions of the DBF file format can specify a codepage.  The 30th byte (`file[29]`) is the "Language Driver ID" and there's a mapping in the codebase just above the function.  When should the file override the option?

dnafigator:
We came across different flies with zero at 30 byte. So, it's possible to
analyse the file and to override the codepage.

SheetJSDev:
Fixed in c56347eff58bc34c393ec35ee27d8a35763847f1

Issue #1300, opened 2018-10-05
Parsing string values and treating as dates is incorrect


pzhelnov:
Hi!
There is a CSV file, which I try to read and it contains field with value

"Aprobil P 0.1%"

the short example of CSV:
```

"Country";"Product Family"
"Germany";"Aprobil P 0.1%"

```

conversion to workbook is the following:

```
var workbook = XLSX.read(csvData, {
                    type:'string',
                    dateNF: 'D-M-YYYY',
                    cellDates:true,
                    cellText:true,
                    cellNF: false,
                raw:false});

```

after conversion I save the XLS, where value "Aprobil P 0.1%" is converted to a date  01.04.00

looking into the worksheet model and getting the certain cell, it contains:
```
{
      t: 'd',
      v: 'Sat Apr 01 2000 00:00:00 GMT+0300 (Eastern European Summer Time)',
      z:undefined
}
```
is there any way to cover this case?
Could you pls assist, what to do.

Thanks in advance!

pzhelnov:
any update, any comments? is anybody here? ;)

metal4timmy:
I am having a similar issue with certain strings being converted to date values.  Any update on this?

pzhelnov:
guys, @SheetJSDev, pls, just pay a little attention to it, tell us, is it possible to fix, or should we rely on our workaround?

pzhelnov:
guys, one year has been passed, now one even commented. 
Do I need to switch to another project?

SheetJSDev:
Hello @pzhelnov ! First, understand that this is an open source project and all work done on the project is charity.  If you are concerned about response time or interested in supporting development, we offer paid builds and support as discussed in https://sheetjs.com/pro 

That said, unfortunately there's no magic solution here.  The Date parsing currently falls back on the JS Date builtin features, which are super flexible in chrome.  As an example:

```js
new Date("This is not a date 1")
```

is the date January 1 2001, despite it clearly indicating that it isn't a date.  

Setting the option `raw: true` will give you the raw strings, suppressing any sort of value interpretation.

If you are interested in helping, the relevant function is [`fuzzydate`](https://github.com/SheetJS/js-xlsx/blob/master/bits/20_jsutils.js#L126).  Google Docs / Office Online likely use a series of regular expressions based on some known formats like `/\d{4}-\d{2}-\d{2}/` for `yyyy-mm-dd` and that might be a more sensible approach than using the generic Date constructor

[...]

Issue #2196, opened 2020-12-11
Unexpected Automatic Date Conversion


matthew-macgregor:
# Issue Overview

One of our systems which uses `sheetjs` for parsing csv/xlsx data produced an unexpected result this week. The data which caused the issue was `"Mayslanding, NJ 08234"`, which the library coerced to a `date` as `{ t: 'n', v: 2313568, w: '5/1/34' }`. My hunch is that the code which attempts to guess if a cell might be a date is too lax given some inputs.

This happens when `raw: false`, which is the default. We have worked around it by simply setting `raw: true` and doing our own type conversions. Still, this unexpected result seems likely to trip up others using the default setting.

I am willing to volunteer to submit a PR to fix this issue but would like some feedback from a maintainer before writing any code. My guess is that any changes to this date conversion code could very easily cause other unexpected side effects, so advice would be appreciated.

### Steps to Reproduce:

Given the (somewhat fabricated) csv data below, the following code will produce the result:

```
const wkbk = xlsx.readFile('./datesville.csv')
```

CSV input:
```csv
Line 1
"Januaryville, NJ 08234"
"Februaryville, NJ 08234"
"Marchville, NJ 08234"
"Aprilville, NJ 08234"
"Mayslanding, NJ 08234"
"Juneville, NJ 08234"
"Julyville, NJ 08234"
"Augustville, NJ 08234"
"Septemberville, NJ 08234"
"Octoberville, NJ 08234"
"Novemberlanding, NJ 08330"
"Decembertown, NJ 08330"
```

Unexpected result:
```js
{
  A1: { t: 's', v: 'Line 1' },
  A2: { t: 'n', v: 2313448, w: '1/1/34' },
  A3: { t: 'n', v: 2313479, w: '2/1/34' },
  A4: { t: 'n', v: 2313507, w: '3/1/34' },
  A5: { t: 'n', v: 2313538, w: '4/1/34' },
  A6: { t: 'n', v: 2313568, w: '5/1/34' },
  A7: { t: 'n', v: 2313599, w: '6/1/34' },
  A8: { t: 'n', v: 2313629, w: '7/1/34' },
  A9: { t: 'n', v: 2313660, w: '8/1/34' },
  A10: { t: 'n', v: 2313691, w: '9/1/34' },
  A11: { t: 'n', v: 2313721, w: '10/1/34' },
  A12: { t: 'n', v: 2348815, w: '11/1/30' },
[...]

Issue #1646, opened 2019-09-30
Invalid date recognition


krojew:
fuzzydate() function incorrectly guesses date due to invalid regex pattern. For example, let's look at "Trollahaugen 64" text. First, it tries to convert it to date and it ends up being 01.01.1964 due to 64 at the end. Then it proceeds with looking for a month using /jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec/ regex. This is the source of error, since the text contains "aug" inside but it's not a separate word, hence should not be considered a month. Yet, the function assumes it is and returns 01.01.1964.

SheetJSDev:
Just so we're on the same page: the root problem is the incredible flexibility of the Date constructor in V8 and other JS engines.  In chrome, `new Date("Trollahaugen 64")` is interpreted as the beginning of the year 1964 (Jan 1 0:00 local time).

There are a handful of workarounds for common cases we've seen but it's not pretty.  The "right" answer is to skip the date constructor entirely and just regex against a series of culture-specific formats

krojew:
I understand the inherent problem with solving such issue - there always will be some edge cases. Working around this particular one seems to be trivial - simply expect the month name to be a whole word, surrounded with non-alphabetic characters, if we assume (which might not be true) that the string is a ECMA-262 Date.toString() representation.

tophsic:
Hi, I had a similar issue with "SEPEM GRENOBLE 2018" text which is transformed in 9/1/18 date.

SheetJSDev:
Moving the discussion to https://github.com/SheetJS/sheetjs/issues/1300

@krojew The spec uses the phrase "implementation-dependent":

> in general, the value produced by Date.parse is implementation-dependent when given any String value that does not conform to the Date Time String Format

That Date Time String Format is related to ISO 8601 and does not cover the set of date expressions understood by Excel

PR #2354, opened 2021-08-13
Issue #1300 fix: fuzzydate strict date option regex

nandanv2702:
Hey! This fix refers to [issue #1300](https://github.com/SheetJS/sheetjs/issues/1300)

I made changes to two of the files to include regex options to parse dates in the following formats:
- mm dd yyyy 
- mm dd yy
- dd mm yyyy
- dd mm yy

All the aforementioned date formats use the separators `. or - or /`, e.g.:
- dd-mm-yy
- dd.mm.yy
- dd/mm/yy

This can be activated using the strictDate option in the fuzzydate function which is set to `false` by default.

Let me know whether this solution is what you were looking for!

### DIFF
--- a/bits/20_jsutils.js
+++ b/bits/20_jsutils.js
@@ -130,8 +130,33 @@ function fuzzynum(s/*:string*/)/*:number*/ {
 	if(!isNaN(v = Number(ss))) return v / wt;
 	return v;
 }
-function fuzzydate(s/*:string*/)/*:Date*/ {
+
+// date regex reference - https://www.oreilly.com/library/view/regular-expressions-cookbook/9781449327453/ch04s04.html
+// matches mm/dd/yy, mm/dd/yyyy, dd/mm/yy, dd/mm/yyyy
+const STRICT_DATE_REGEX = [
+	"^(?:(1[0-2]|0[1-9])\\.(3[01]|[12][0-9]|0[1-9])|(3[01]|[12][0-9]|0[1-9])\\.(1[0-2]|0[1-9]))\\.([0-2][0-9]{3}|[0-9]{2})$", 
+	"^(?:(1[0-2]|0[1-9])\/(3[01]|[12][0-9]|0[1-9])|(3[01]|[12][0-9]|0[1-9])\/(1[0-2]|0[1-9]))\/([0-2][0-9]{3}|[0-9]{2})$",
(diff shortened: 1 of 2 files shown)


SheetJSDev:
ES6 issues aside (we are targeting ES3, so no const or default parameter values or arrow functions) two comments:

1) the result of the match is not used, so [RegExp#test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) probably should be used.  You might also be missing a new RegExp somewhere.

2) how should the library interpret something ambiguous like `03-02-01` or `01/02/03` ?

nandanv2702:
Hey! I'll work on the changes for the first point this weekend, but as for the second - what do you think of including an option similar to `dateNF` that allows a user to enter the date format as `dd-mm-yy, mm-dd-yy` etc.? 

Alternatively, if `strictDates` is set to true, we can use `dateNF` (or the newly defined similar variable as mentioned above) to match any date values to the regex options defined in this PR [here](https://github.com/SheetJS/sheetjs/pull/2354/files#diff-4aeb62dc7a6f86c9dedff5f3d3e3144d5a0269b3d9ed7e5cf50d9b39f2a36122R136)

What do you think?

nandanv2702:
Just following up on my last comment - after reading #2196 and #1646, I think it might make more sense to make the following changes:

1. Change `strictDates` to a string that can be `dd-mm`, `mm-dd`, or `true`
2. Refactor [`STRICT_DATE_REGEX`](https://github.com/SheetJS/sheetjs/pull/2354/files#diff-4aeb62dc7a6f86c9dedff5f3d3e3144d5a0269b3d9ed7e5cf50d9b39f2a36122R136) to be an Object with the keys `dd-mm` and `mm-dd with their respective regex values

This can allow for stricter dates while also letting us switch `dd-mm` to `mm-dd` internally since the JS Date function only uses `mm-dd`. In case `strictDates` is `true`, we can test it with all [`STRICT_DATE_REGEX`](https://github.com/SheetJS/sheetjs/pull/2354/files#diff-4aeb62dc7a6f86c9dedff5f3d3e3144d5a0269b3d9ed7e5cf50d9b39f2a36122R136) values and return the first match (which will, by nature, be in the `mm-dd` format).

I'm currently working on these changes, so let me know if this sounds good!

PR #2383, opened 2021-09-13
Add check for finite number

### LINKED ISSUE #2373: Some strings returns INIFINITY
Settings applied: **raw: true, rawNumbers: true**
Example of string wrong interpreted:
"3E17773446"

The whole row in CSV was:
"86","3E17773446","For sale ","5/5/2021 11:12:18 AM"

KevinBatdorf:
I implemented the change on the issue but not clear on whether to include the build files in the PR. I ran a few of the make commands ([also not super clear how that works to a newcomer](https://github.com/SheetJS/sheetjs#osxlinux)), and it modified around 15 files. I can push those changes here as well, so just let me know. 

I'm also not sure on testing in general. Does this function have test coverage? I noticed the make command pulled in a submodule, but I didn't dive in so it could be I missed something.

Finally, I'm also not sure how to manually test it. I saw the index.html but I thought it may have brought up a web interface. Is the idea to create a spreadsheet with the reproduction, then upload it/ drag&drop there?

Thanks!

Closes #2373

### DIFF
--- a/bits/20_jsutils.js
+++ b/bits/20_jsutils.js
@@ -121,7 +121,8 @@ function fill(c/*:string*/,l/*:number*/)/*:string*/ { var o = ""; while(o.length
 /* TODO: stress test */
 function fuzzynum(s/*:string*/)/*:number*/ {
 	var v/*:number*/ = Number(s);
-	if(!isNaN(v)) return v;
+	if(!isFinite(v)) return v;
+	if(!isNaN(v)) return NaN;
 	if(!/\d/.test(s)) return v;
 	var wt = 1;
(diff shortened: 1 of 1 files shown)


KevinBatdorf:
Not sure about the failed workflow either. Can follow up during the week though!

SheetJSDev:
Failed workflow and failed builds are stemming from some npm issue with certain older versions of node.  There used to be some workarounds for Travis CI, but they didn't neatly map to the new github actions.  If there are actual issues, they will show up in local testing before the next release.

SheetJSDev:
e7346ddfbe4ae40d8193e558cc61dc7eb6ea9511

PR #2405, opened 2021-10-02
Use named define for requireJS

gfyoung:
Two benefits:
    
* Helps avoid with mismatched anonymous errors

ref: https://requirejs.org/docs/errors.html#mismatch
 
* Allows for uglification in the demo


SheetJSDev:
Should it be `'xlsx-dist'` or `'xlsx'`?

gfyoung:
> Should it be 'xlsx-dist' or 'xlsx'?

`xlsx-dist` worked fine when I tried running the demo.

gfyoung:
https://github.com/SheetJS/sheetjs/actions/runs/1299192977

FYI: there was no option for me to select this.

Issue #2410, opened 2021-10-06
Update x-spreadsheet demo with improved import/export funtions


ThibautSF:
Hi,

Made improvements on export and import functions between SheetJS and [x-spreadsheet](https://github.com/myliang/x-spreadsheet/).
You can find more details for these improvements on the following "issue" in x-spreadsheet GitHub: myliang/x-spreadsheet#419.

PR #2411, opened 2021-10-06
Improve import/export functions for sheetjs with x-spreadsheet

ThibautSF:
Pull Request for #2410

Integration of improves import and export functions in x-spreadsheet demo from myliang/x-spreadsheet#419.

Allow optional arguments to support new QoL functionalities (default to previous behaviour)
- keep merged cells between formats
- choose to keep formulas (as string ie '=SUM(...)') or formula result (default behaviour)

Note: Incorporate suggestions by @Liyaa

### DIFF
--- a/demos/xspreadsheet/xlsxspread.js
+++ b/demos/xspreadsheet/xlsxspread.js
@@ -164,6 +164,6 @@ function xtos(sdata, keepMerges, keepFormulas) {
 
         XLSX.utils.book_append_sheet(out, ws, xws.name);
     });
+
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Thanks for looking into this!

Is there a reason not to enable keepMerges?  (I think the option should be removed and it should always keep merges)

Is there a way to distinguish the text `=1+1` from the formula `=1+1`?  If not, the formula logic is fine.

Ideally we'd merge one commit, so please squash down

PR #2420, opened 2021-10-15
Use requirejs 2.3.6 in requirejs demo

gfyoung:
Just using the latest and greatest

xref: https://github.com/SheetJS/sheetjs/pull/2405


reviewher:
a7405c4

PR #2423, opened 2021-10-17
Fix CSV newlines to not add an extra empty row

### LINKED ISSUE #1553: Converting file to csv stream is adding new line
I am trying to convert the xlsx file to csv stream and when i listen on `data` and print the chuck i'm seeing a new empty line which is not present in the file

```js
var xlsx = require('xlsx')
const wb = xlsx.readFile('testData.xlsx')
const ws = wb.Sheets[wb.SheetNames[0]];
const stream = xlsx.stream.to_csv(ws)
stream.on('data', chunk => {
    console.log(chunk)
    console.log(chunk.toString())
    console.log(chunk.toString().length)
})
```

the `testData.xlsx` file contains the following data
```
a,b,c
1,2,3
```
When i run the above mentioned code i'm getting the following output
```
<Buffer ef bb bf>

1
<Buffer 61 2c 62 2c 63 0a>
a,b,c

6
<Buffer 31 2c 32 2c 33 0a>
1,2,3

6
```
in which we can see a new line.

Can someone help me understand why xlsx is adding that new line and how to read the correct data from the file?

TheGLander:
Fixes #1553

Does the suggested fix, maybe I should add a test case?

### DIFF
--- a/bits/97_node.js
+++ b/bits/97_node.js
@@ -16,6 +16,7 @@ if(has_buf && typeof require != 'undefined') (function() {
 		for(var C = r.s.c; C <= r.e.c; ++C) if (!((colinfo[C]||{}).hidden)) cols[C] = encode_col(C);
 		var R = r.s.r;
 		var BOM = false;
+		var wrote_first_line = false;
 		stream._read = function() {
 			if(!BOM) { BOM = true; return stream.push("\uFEFF"); }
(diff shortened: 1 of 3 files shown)


SheetJSDev:
Fixed in a32b30414b3a1da61ba6d6b847a473f54481815a . CSV export (both stream and normal) omits trailing record separator

PR #2459, opened 2021-11-23
fix: for XLSX.stream.to_json doesn't ends when a blank row is the las…

rohandhamapurkar:
@SheetJSDev sorry was pretty late on this pull request but the change that you have suggested in #1779  I have made a pull request for the same as it fixes the issue for me, tested this in production.


SheetJSDev:
467020fc693f1c34cfa32670022c3f8ef4199a58

PR #2461, opened 2021-11-24
fix issue #2460

giuliohome:
See the [issue](https://github.com/SheetJS/sheetjs/issues/2460) for more details

Issue #2460, opened 2021-11-24
Wrong order for undefined cells in the first row in case of date types


giuliohome:
Afaics, date columns are shifted to the end (when they are empty on the first row under the headers) because the json order is the property inserts, but the undefined header value is skipped in `function make_json_row` [here](https://github.com/SheetJS/sheetjs/blob/master/bits/90_utils.js#L19), hence the column is incorrectly reordered to the end when the value is finally inserted the first time in the next rows.

I'm going to **try** this change, locally on my side and I'll create a PR after my test succeeds.

```javascript
		if(val === undefined || val.t === undefined) {
			if(defval === undefined) {
				row[hdr[C]] = null; // <-- this is my proposed solution to the issue
				continue;
			}
			if(hdr[C] != null) { row[hdr[C]] = defval; }
			continue;
		}
```

(notice that this change seems to break old tests about json output, because it introduces null fields)

giuliohome:
Solved with native option `defval: null` in `sheet_to_json` (no PR needed)

SeniorChiefABLEUnit:
Confirmed/no pr/R-D EEngineer is choice/no money for spending my entire
life instructing others positively for Lightning Locker Service will
require you motherfuckers to go to bed at night with a clear conscience. My
conscience is clear, I will NOT change who I am for anybody.
-Bryan Bellipanni
November 24,2021

giuliohome:
Actually, if this is only for ordering purpose, it could make sense to introduce a different option, e.g. `keeporder: true`, that _somehow_ corresponds to `defval: null` but only for the **first row**. In this sense, the issue can be reopened and a PR could be made accordingly.

(btw, I've blocked the above spammer)

giuliohome:
Merged the two commits into one and PR reopened: the final code is

```javascript
if(defval === undefined) { 
	if (R == 1 && o.keeporder) row[hdr[C]] = null; 
	continue; 
}
```

as described in my last comment.

SheetJSDev:
Can you actually share a case when this happens?  To be clear, the ordering when calling `XLSX.utils.sheet_to_json` is determined before scanning the first data row: https://github.com/SheetJS/sheetjs/blob/master/bits/90_utils.js#L69

Suppose you start with:

```js
var ws = XLSX.utils.aoa_to_sheet([
  ["a", ,"c"], // second header missing
  [1, 2, 3],
  [4, 5, 6]
]);
```

The actual behavior depends on the `header` parameter:

[...]

PR #2462, opened 2021-11-24
issue #2460

giuliohome:
The PR #2461 has been redone here to pass the automatic rebase test.
In case this PR is approved and merged, I can prepare a follow-up change for the [json options](https://github.com/SheetJS/sheetjs#json) documentation, adding a line in the table for the new `keeporder` option, very similarly to the existing `defvalue`.


SheetJSDev:
See comment in #2460

PR #2467, opened 2021-12-03
[README.md] Fixed error in parameter name

jeijei4:
Fixed error in parameter name.
=> Browser file upload form element (Using :Blob#arrayBuffer)


SheetJSDev:
5b25a430f2e33516092fedf7f0d10d494d4df9c5

PR #2477, opened 2021-12-16
avoid utf-8 buffer re-decoding

hugoaboud:
Related to [this issue](https://github.com/SheetJS/sheetjs/issues/2476), fix suggested by @SheetJSDev.


Yogu:
Great! I think this condition should be added to the first `if` clause, too, so UTF-8-encoded buffers starting with a BOM also are not decoded twice.

reviewher:
92d8a38

PR #2493, opened 2022-01-17
Fix types used by table_to_book()

michaelmarziani:
This fix allows naming the sheet and using other writing options.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -205,8 +205,16 @@ export interface ParsingOptions extends CommonOptions {
     dense?: boolean;
 }
 
+export interface SheetOption {
+  /**
+   * Name of Worksheet (for single-sheet formats)
+   * @default ''
+   */
+  sheet?: string;
+}
+
 /** Options for write and writeFile */
-export interface WritingOptions extends CommonOptions {
+export interface WritingOptions extends CommonOptions, SheetOption {
     /** Output data encoding */
     type?: 'base64' | 'binary' | 'buffer' | 'file' | 'array' | 'string';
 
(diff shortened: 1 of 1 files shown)


michaelmarziani:
History of this PR:

I wrote this code in React:

```typescript
const handleExportToExcel = () => {
  const table = document.getElementById('data-table');
  var workBook = XLSX.utils.table_to_book(table, { sheet: "Sheet JS" });
  return XLSX.writeFile(workBook, 'SheetJSTableExport.xlsx');
};
```

TypeScript gave an error:
_Argument of type '{ sheet: string; }' is not assignable to parameter of type 'Table2SheetOpts'.
  Object literal may only specify known properties, and 'sheet' does not exist in type 'Table2SheetOpts'.ts(2345)_

I noticed in the code that the `WritingOptions` interface extends `CommonOptions` and has sheet declared inside it. When I changed `CommonOptions` to `WritingOptions` in the interface declaration for `Table2SheetOpts`, I no longer get this TS error and the function works as expected for my use case.

SheetJSDev:
The other writing options are not supported in `table_to_book` or `table_to_sheet`.  Can you instead introduce another interface `SheetOption`?  Then `Table2SheetOpts` and `WritingOptions` should extend `SheetOption`

michaelmarziani:
I can do that. Thanks for the suggestion.

michaelmarziani:
@SheetJSDev I believe I have done as you requested. Please let me know how it looks.

michaelmarziani:
@SheetJSDev I just wanted to follow up on this PR, please let me know if it needs anything else at your convenience.

michaelmarziani:
@SheetJSDev Thank you!

Issue #2485, opened 2021-12-30
problem with numeric formatting with CSV file.


t-heu:
how can i leave all values in string? I have some numeric values that are coming out broken for example: `297092.999351851` (supposed to be this result: `05.030.2713`) and `17508` (supposed to be this result: `175.08`). This only happens with CSV file upload.

Version xlsx: 0.17.2

steviegalluscio:
I was debugging this issue and found that you can set the raw property to true and it stores the values as strings without attempting to process them.

`const wb = XLSX.read(content, { type: 'buffer', raw: true });` 

I tested it out with the values you gave and this returns your expected results.

t-heu:
Thank you

Issue #1636, opened 2019-09-21
[0.15.1] Text being converted to dates


Webifi:
When exporting a table with table_to_book, a cell with a text value of "Smith, Julie + 5" is being converted to "07/05/2001".  {raw: true} fixes this, but then columns with actual dates are not being converted. 

Edit:
Looks like the issue is with the fuzzydate method.

While it won't fix everything (fuzzydate is way too, well, fuzzy), changing:

`if(s.toLowerCase().match(/jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec/)) return o;`

to:

`if(s.match(/\b(jan(\b|uary\b)|feb(\b|ruary\b)|mar(\b|ch\b)|apr(\b|il\b)|may\b|jun(\b|e\b)|jul(\b|y\b)|aug(\b|ust\b)|sep(\b|tember\b)|oct(\b|ober\b)|nov(\b|ember\b)|dec(\b|ember\b))(\s*|\/|-)[0-9]+/i)) return o;`
`if(s.match(/[0-9]+(\b|st|th|\/|-)\s*\b(jan(\b|uary\b)|feb(\b|ruary\b)|mar(\b|ch\b)|apr(\b|il\b)|may\b|jun(\b|e\b)|jul(\b|y\b)|aug(\b|ust\b)|sep(\b|tember\b)|oct(\b|ober\b)|nov(\b|ember\b)|dec(\b|ember\b))/i)) return o;`

would at least make the month match more restrictive.

Currently it looks like fuzzydate would match a date on text like "... You **may** need 5 ..." or "**jul**ie is missing 5 teeth" or "6 children **mar**ching" or even "I fear l**aug**ing clowns".

The updated regexps would require "may 5" or "jul 5" or "july 5th" or "6 march"  (abbreviated or full month followed by a number or number (+optional ordinal) followed by month)

CNzxy0820:
May I request a question , whether the js-xlsx can show the excel datas use the format of "yyyy.m.d h:mm" not the  "yyyy.m.d h:mm" to text format,when I used the js-xlsx , the "2019.9.22 16:22" was convert to "43730.682291666664",I want to add the datetime like "2019.9.22 16:22" to as ECharts'xaxis. I am sorry my english is poor ,very sorry...  if you can answer my ,I will very very happy!!

Webifi:
[...]

PR #2502, opened 2022-01-24
More restrictive fuzzydate

steviegalluscio:
This change makes the fuzzydate function more restrictive. The changes are:

- For dates without a month name, only dates with **two** valid separators(`/` `-` `.`) are allowed.
ex. `1/2` is no longer assumed to be a date where as `1/2/0` is acceptable. 
dates with spaces as a separator are not assumed to be dates such as `1 2 0` 
Note: `1 / 2 / 0` is acceptable
Also fixes `AUD 76.09` from a [comment ](https://github.com/SheetJS/sheetjs/issues/1636#issuecomment-557643678) on issue #1636  will no longer be parsed as a date.

- Values such as `05.030.2713` will no longer be accepted as dates because of the month having 3 digits (a total of > 10 digits) #2485 

I have seen many issues created over fuzzydate and while raw parsing is a workaround, many people want other numeric values preserved and by default. This is a proposed solution by dropping support for mm/dd, mm/yyyy, mm.yyyy etc. 

I assume support for dates with a space separator is necessary so let me know if it's worth continuing in this direction. An issue not addressed by this solution is the automatic date conversion of values with two decimal points such as: `3.5.2`. Values like these are often used for software versioning. I would prefer not to have to use decimal separator with dates as it's much less common but was curious if any other spreadsheet software uses this standard.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -2873,11 +2873,16 @@ function fuzzydate(s) {
 	var o = new Date(s), n = new Date(NaN);
 	var y = o.getYear(), m = o.getMonth(), d = o.getDate();
 	if(isNaN(d)) return n;
-	if(y < 0 || y > 8099) return n;
-	if((m > 0 || d > 1) && y != 101) return o;
-	if(s.toLowerCase().match(/jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec/)) return o;
-	if(s.match(/[^-0-9:,\/\\]/)) return n;
-	return o;
+	if(s.toLowerCase().match(/jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec/)){
+		if(y < 0 || y > 8099) return n;
+		if((m > 0 || d > 1) && y != 101) return o;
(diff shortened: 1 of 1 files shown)


reviewher:
Please leave feedback in the issue #1300 first (explaining in words what you think is an acceptable date)

SheetJSDev:
There are some interesting ideas here, but it seems to invalidate common Date+Time strings.  For example, consider the date time string `02/02/2022 02:02:02 AM`.  This is a valid V8 date:

```js
new Date("02/02/2022 02:02:02 AM");
// Wed Feb 02 2022 02:02:02 GMT-0500 (Eastern Standard Time)
```

This is rejected by the string length test:

```js
 "02/02/2022 02:02:02 AM".replace(/\s+/g, "").length
// 20
```

Excel recognizes `02/02/02 02:02:02.222 AM` as a valid datetime (it assumes the format mm:ss.0 by default, but you can change the number format to something like `yyyy-mm-dd hh:mm:ss.000` to see that it extracted the correct value)

PR #2528, opened 2022-02-20
NestJS/SheetJS demo instructions

dmlo:
The demo creates a NestJS server that uses SheetJS to import an uploaded spreadsheet.

### DIFF
--- a/demos/server/README.md
+++ b/demos/server/README.md
@@ -163,4 +163,96 @@ curl -X POST http://localhost:7262/file?f=sheetjs.csv
 curl -X GET http://localhost:7262/?f=sheetjs.xlsb
 ```
 
+
+
+## NestJS
+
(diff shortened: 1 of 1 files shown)


SheetJSDev:
Thanks for looking into this!  Few comments:

1) Add a sample cURL line for testing, like 

```
curl -X POST -F "file=@test.xlsx" http://localhost:3000/sheetjs/upload-xlsx-file
```

(is 3000 the default port for NestJS?)

2) According to `nest new -h`, "Select NPM as the project's package manager when prompted." can be effectuated non-interactively with a CLI flag:

```bash
nest new -p npm xlsx-demo
```

reviewher:
56e23ca8e09e5513a8214c9d39d11a0366353daf

Issue #2505, opened 2022-01-24
Reading workbook with embedded word document may read the embedded doc props instead of the xls props


lionel87:
We have an old xls book, on the sheet there is a logo which comes from an embedded word document.
Reading the props of this workbook (Author, CreatedAt etc.) gives the embedded word document props instead of the xls workbook props.

The code:
```js
console.log(require('xlsx').readFile('k061006.xls').Props);
```
This shows eg. the author of the embedded document instead of the author of the xls file.

Attached the file which produces this bug: [k061006.xls](https://github.com/SheetJS/sheetjs/files/7926479/k061006.xls)

Used package version: xlsx 0.17.5

SheetJSDev:
Fortunately this is a two-line fix to `bits/76_xls.js`, feel free to send a PR:

```diff
--- a/bits/76_xls.js
+++ b/bits/76_xls.js
@@ -879,14 +879,14 @@ var PSCLSID = {
 };
 function parse_xls_props(cfb/*:CFBContainer*/, props, o) {
        /* [MS-OSHARED] 2.3.3.2.2 Document Summary Information Property Set */
-       var DSI = CFB.find(cfb, '!DocumentSummaryInformation');
+       var DSI = CFB.find(cfb, '/!DocumentSummaryInformation');
        if(DSI && DSI.size > 0) try {
                var DocSummary = parse_PropertySetStream(DSI, DocSummaryPIDDSI, PSCLSID.DSI);
                for(var d in DocSummary) props[d] = DocSummary[d];
        } catch(e) {if(o.WTF) throw e;/* empty */}
 
        /* [MS-OSHARED] 2.3.3.2.1 Summary Information Property Set*/
-       var SI = CFB.find(cfb, '!SummaryInformation');
+       var SI = CFB.find(cfb, '/!SummaryInformation');
        if(SI && SI.size > 0) try {
                var Summary = parse_PropertySetStream(SI, SummaryPIDSI, PSCLSID.SI);
                for(var s in Summary) if(props[s] == null) props[s] = Summary[s];
```

lionel87:
Thanks for the fast reply. Your solution worked for us, sent a PR.

reviewher:
Fixed in d01661565afb6df5a9c7d56a957df79ac5efcd9c

PR #2544, opened 2022-03-06
Fix #2543

### LINKED ISSUE #2543: Error: Invalid HTML: could not find <table>
```
//  index.js

const XLSX = require("xlsx");
const workbook = XLSX.readFile("test.html");
```
I am running into this error unexpectedly.
Can't figure out the reason why.

The html does have proper ```<table>``` tags for the only table in the file.

HritikR:
Fixes issue #2505. Fix suggested by @SheetJSDev


reviewher:
a32b30414b3a1da61ba6d6b847a473f54481815a

PR #2657, opened 2022-04-18
Update README.md

edulecca:
- Type definition differs from README
- tested and works as tyoe defines

```
type CBFunc = () => void;
export function writeFileAsync(filename: string, data: WorkBook, opts: WritingOptions | CBFunc, cb?: CBFunc): any;
```


SheetJSDev:
90747905ad56acbcd93f675deaf239e0570fbe25

PR #2670, opened 2022-04-30
Inline js task

calcscout:
Added additional check when not to create a link in a cell. Now if <a> tag has "href" with inline js, hyperlink is not created.


SheetJSDev:
f38191d266c91ad73d7db3c907e42fa808302b67

PR #2683, opened 2022-05-13
Fix basedate calculation

ivan-trusov:
Current method of numdate/datenum calculation works incorrectly for timezones that are equal nowadays but were different in the past (e.g. Europe/Minsk and Europe/Moscow). Suggested method bypasses this issue by initializing basedate to UTC 1899-11-30 00:00:00 and slightly modifiing TZ offsets handling.

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -3214,16 +3214,19 @@ function evert_arr(obj) {
 	return o;
 }
 
-var basedate = new Date(1899, 11, 30, 0, 0, 0); // 2209161600000
+var basedate = new Date(); // 2209161600000
+basedate.setUTCFullYear(1899, 11, 30);
+basedate.setUTCHours(0, 0, 0);
+
 function datenum(v, date1904) {
 	var epoch = v.getTime();
 	if(date1904) epoch -= 1462*24*60*60*1000;
-	var dnthresh = basedate.getTime() + (v.getTimezoneOffset() - basedate.getTimezoneOffset()) * 60000;
+	var dnthresh = basedate.getTime() + v.getTimezoneOffset() * 60000;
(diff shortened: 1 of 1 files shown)


xlsx.js
```diff
@@ -3214,16 +3214,19 @@ function evert_arr(obj) {
 	return o;
 }
 
-var basedate = new Date(1899, 11, 30, 0, 0, 0); // 2209161600000
+var basedate = new Date(); // 2209161600000
+basedate.setUTCFullYear(1899, 11, 30);
+basedate.setUTCHours(0, 0, 0);
```
Fil, line 3219:
```suggestion
var basedate = new Date(Date.UTC(1899, 11, 30, 0, 0, 0)); // 2209161600000
```

ivan-trusov: Good idea. Thanks

SheetJSDev:
We'll revisit dates later this month.  Please don't merge with latest, as the actual commits can be cherrypicked as needed.

Given the [4-year old V8/NodeJS/Chrome timezone bug](https://bugs.chromium.org/p/v8/issues/detail?id=7863) and the declining prospects of a real fix to V8, we'll have to adjust.  Would it make more sense to anchor to a date that does not suffer from the bug?  Looking at the tz database, the last timezone to align to integral minutes is `Africa/Monrovia`, which shifted on 1972-01-07 (date code 26305 in the 1900 date system)

ivan-trusov:
>  Please don't merge with latest, as the actual commits can be cherrypicked as needed.
My fail. Apologize.
I slightly refactored numdate()/datenum() functions. AFAIK OADate must be treated as it has no TZ or DST issues ("always in local TZ"), so I suggest to do all calculations in UTC and at the very end de-adjust TZ offset forcibly added by JS engine

PR #2691, opened 2022-05-22
Export choose_fmt

alonkh2:
in order to implement format with limited width as described here (https://github.com/SheetJS/sheetjs/issues/2689) externally we need to use "choose_fmt" and "isgeneral".


SheetJSDev:
FWIW `isgeneral` quite literally checks for the string "General".

PR #2693, opened 2022-05-24
Update Application property

purusothsv13:
Update Workbook property only if it is not already set in the Workbook object.


SheetJSDev:
It's done this way to prevent roundtripping an Application string of "Microsoft Excel" or "Microsoft Macintosh Excel" (since SheetJS is not actually Excel).  If you really want to override, use a Proxy like in https://github.com/SheetJS/sheetjs/issues/1231#issuecomment-415192707

PR #2705, opened 2022-06-07
feat: update `Sheet` type

bluelovers:


SheetJSDev:
What TS version introduced that syntax?  The main concern is not breaking the TS versions tied to older versions of angular.  Angular 4.x depended on TS 2.2

SheetJSDev:
This doesn't even appear to work in the latest TS version: https://www.typescriptlang.org/play?#code/G4QwTgBAHgXBAGASA3gZwC5gJYDsDmAvijgK4C2ARgKZgHwQC8EARAIIBCAwgIwBMAzMwDcAKCA

```typescript
var x: `${string}${number}` = "ABC123";
```

> Type '"ABC123"' is not assignable to type '`${string}${number}`'.

It would be neat if there were a way to specify a RegExp for string literal unions but alas.

bluelovers:
@SheetJSDev this not work for every cell key, but it still is help

Issue #2560, opened 2018-01-09
hh:mm:ss formatting bug


phillipperalez:
when time is of a value that should round up the minute, if the second has a u (rounding value I assume) greater than or equal to .5, then the seconds will go to `00` without bumping the minute value up one. In excel the time is properly rounded

In this example I am using ` require('make-runnable') ` at the end of ` ssf.js ` to test this via CLI. The value is from 119.99 / 86400 - so the result I expect should be 00:02:00, however I receive 00:01:00 instead

` node ssf.js format hh:mm:ss 0.001388773 `

PR #2717, opened 2022-06-15
fix date/time rounding issue

scottysseus:
See: #2560

- Improved rounding to account for minutes, hours, etc.
- Added tests for rounding dates/times

### DIFF
--- a/packages/ssf/bits/35_datecode.js
+++ b/packages/ssf/bits/35_datecode.js
@@ -1,5 +1,6 @@
 function parse_date_code(v/*:number*/,opts/*:?any*/,b2/*:?boolean*/) {
 	if(v > 2958465 || v < 0) return null;
+	opts.b2 = b2 || false;
 	var date = (v|0), time = Math.floor(86400 * (v - date)), dow=0;
 	var dout=[];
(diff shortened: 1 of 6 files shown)


SheetJSDev:
The tests predictably failed on old node versions since they did not support ES6 destructuring.

Did you test against the examples provided by @snoopyjc in https://github.com/SheetJS/sheetjs/issues/2560#issuecomment-1062838050 ?

scottysseus:
@SheetJSDev I would like to get this change merged if there aren't any other issues and you think it's appropriate. Is there anything else I need to do?

SheetJSDev:
Nothing else on your end.  There are some logistical issues we are figuring out regarding date handling.

Ideally, 0.19.0 resolves the outstanding date processing problems which requires both the SSF formatting library and the main file processing library to agree on date code interpretation.

MrToadie:
Thanks for all your work!

scottysseus:
Are there any updates on this? At a quick glance, I'm not able to reproduce this here: https://oss.sheetjs.com/ssf/

but want to make sure this has been fixed and what version it's been fixed in

PR #2744, opened 2022-07-15
add exports field to package.json

BrianHung:


SheetJSDev:
What tools or platforms are affected by the change?

BrianHung:
Running into an issue using SheetJS with vite: https://github.com/vitejs/vite/discussions/9149
Vite is using the CJS distribution of SheetJS and is incorrectly converting that to ESM. This config change will allow Vite to use the ESM distribution of SheetJS directly.

SheetJSDev:
The last time we seriously looked into modules / ESM support, tooling looked for the `module` field in package.json.  This sufficed:

```json
	"main": "xlsx.js",
	"module": "xlsx.mjs",
```

If Vite requires something extra, changes can be made as long as they don't break older NodeJS versions or other bundlers.

There are two changes here: `type: "module"` and `exports`.

[Based on the NodeJS docs](https://nodejs.org/api/packages.html#type), `type: "module"` affects interpretations of `.js` files, but there is no confusion here: the `.js` files are CommonJS and the `.mjs` files are ESM.

With the goal of persuading Vite to use `xlsx.mjs` when importing, would the `exports` field suffice (is `"type": "module"` also needed)?

BrianHung:
I got that off the Vite docs which recommends the following for libraries:

https://vitejs.dev/guide/build.html#library-mode

```js
{
  "name": "my-lib",
  "type": "module",
  "files": ["dist"],
  "main": "./dist/my-lib.umd.cjs",
  "module": "./dist/my-lib.js",
  "exports": {
    ".": {
      "import": "./dist/my-lib.js",
      "require": "./dist/my-lib.umd.cjs"
    }
  }
}
```

After reading the NodeJS docs, `type: module` isn't useful here if the source code isn't written in esm and if it isn't meant to be directly used (?). I'll update the PR.

SheetJSDev:
We pushed 0.18.10 (mainly to make sure that the package.json changes worked)

Does it work in your testing with Vite?

BrianHung:
> We pushed 0.18.10 (mainly to make sure that the package.json changes worked)
> 
> 
> 
> Does it work in your testing with Vite?

Yep, thanks for the quick resolution 🤗.

PR #2763, opened 2022-08-08
fix(read): allow reading base64 files from a dataURI scheme

### LINKED ISSUE #2762: [Improvement] reading excel file from base64 should allow ignoring data uri information
When reading a base64 file, it would be nice if XLSX would ignore the leading data URI parameters.
So it can treat 
`data:text/csv;base64,TmFtZXMNCkhhZmV6DQpTYW0NCg==`
the same way it treats
`TmFtZXMNCkhhZmV6DQpTYW0NCg==`

```
const workBook = XLSX.read(sheetInBase64, { type: 'base64' });
// workBook should be the same whether `sheetInBase64` starts with a dataURI scheme or without it.
```

I would like to create a PR with that fix, but the tests file seem to be automatically generated, and when I run `npm test` I simply get `bits/some_file.js` in the log, with no output telling me whether the tests passed or failed.

AbdelrahmanHafez:
fixes #2762

I'm not if the modified files are source files that get compiled or those are the compiled files.
Also, can you point me to where should the test cases for those be?

```js
it('handles base64 within data URI scheme (gh-2762), function() {
  // Arrange
  const fileInBase64 = 'TmFtZXMNCkhhZmV6DQpTYW0NCg==';
  const fileInBase64WithDataURIScheme = 'data:text/csv;base64,TmFtZXMNCkhhZmV6DQpTYW0NCg==';

  // Act
  const workBookFromRawBase64 = XLSX.read(fileInBase64, { type: 'base64' });
  const workBookFromBase64WithinDataURI = XLSX.read(fileInBase64WithDataURIScheme, { type: 'base64' });

  // Assert
  assert.deepStrictEqual(workBookFromRawBase64, workBookFromBase64WithinDataURI);
});
```

### DIFF
--- a/bits/04_base64.js
+++ b/bits/04_base64.js
@@ -48,7 +48,9 @@ function Base64_encode_pass(input) {
 function Base64_decode(input) {
   var o = "";
   var c1 = 0, c2 = 0, c3 = 0, e1 = 0, e2 = 0, e3 = 0, e4 = 0;
-  input = input.replace(/[^\w\+\/\=]/g, "");
+  input = input.replace(/^data:([^\/]+\/[^\/]+)?;base64,/, '')
+               .replace(/[^\w\+\/\=]/g, "");
+
   for (var i = 0; i < input.length; ) {
     e1 = Base64_map.indexOf(input.charAt(i++));
     e2 = Base64_map.indexOf(input.charAt(i++));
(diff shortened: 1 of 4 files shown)


SheetJSDev:
The test should be added between line 690 and 691 in `test.js`: https://github.com/SheetJS/sheetjs/blob/master/test.js#L690-L691 (just after the "should read base64 strings" test)

The media type is optional and the comma need not be escaped.  Try `/^data:([^\/]+\/[^\/]+)?;base64,/`

Include a test with a string like

```js
  var fileInBase64WithDataURISchemeNoMediaType = 'data:;base64,TmFtZXMNCkhhZmV6DQpTYW0NCg==';
```

The test snippet should use avoid const and any other ES6+ features.

AbdelrahmanHafez:
@SheetJSDev 
Thanks a lot for the detailed comment, I've pushed the necessary changes.

Issue #2673, opened 2022-05-05
More complex headers can cause export style exceptions.


daiki-ver:
Our company requires some information to be added to the header of the exported table. After I add it, the echo in layui is normal, but it is messed up when exporting excel.
![image](https://user-images.githubusercontent.com/71960981/166893361-73425098-8bd0-42df-911d-f29f71584609.png)
![image](https://user-images.githubusercontent.com/71960981/166893290-5be4bc84-74b7-44a9-87a0-7c921b52f369.png)

SheetJSDev:
How is the HTML table being generated?  Can you make that available somewhere?

PR #2764, opened 2022-08-08
fix tests re #2763 #2762

AbdelrahmanHafez:
#2673 was merged with a few failing tests due to a `v` value in the sheets being added to random rows, causing the tests to fail.

So to get consistent results, I am converting the sheet to json, and asserting against that JSON value.
Let's wait and see if all the tests pass on the pipeline, then we can merge.


SheetJSDev:
Simplified the test a bit (reused the base64 string and added the `application/vnd.ms-excel` type) and updated the other tests.  The browser tests also pass https://app.saucelabs.com/builds/vdc/d0856e046ac93728ab5734c276f0ca89

AbdelrahmanHafez:
Thanks @SheetJSDev 👍

PR #2771, opened 2022-08-17
feat(98_esmxport): add esm default export

wkstudy:
Recently I upgraded the xlsx version of the project from 0.15.6=>0.18.5, which caused the project to report an error, and I found that the previous project used the following syntax
```
import xlsx from 'xlsx';
...
xlsx.read(...)
...
```
This upgrade causes the failure to feel still not very good, should you consider adding a default export?
Thanks for the answer

### DIFF
--- a/misc/98_esmxport.js
+++ b/misc/98_esmxport.js
@@ -18,3 +18,22 @@ export {
 	SSF,
 	CFB
 };
+export default {
+	parse_xlscfb,
+	parse_zip,
+	readSync as read,
+	readFileSync as readFile,
+	readFileSync,
+	writeSync as write,
+	writeFileSync as writeFile,
+	writeFileSync,
+	writeFileAsync,
+	writeSyncXLSX as writeXLSX,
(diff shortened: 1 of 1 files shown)


danger2520 (approved):

SheetJSDev:
There's one upgrade caveat that needs to be noted: https://docs.sheetjs.com/docs/getting-started/installation/frameworks#xls-support

The CJS build includes the codepage dependency.  This worked in the past when the library only shipped with the NodeJS and browser standalone builds.  Webpack and other bundlers used the browser build without question.

ESM is a messy ecosystem.  Thanks to incompatibilities between Node ESM / Browser ESM / Deno ESM / "Bundler" ESM, there was no universal way to import the codepage library from the main script.

The uniform solution is to invert the flow.  The ESM builds do not import any dependencies by default, and you now must manually import the codepage tables for XLS support:

```js
import * as xlsx from "xlsx";
import * as cptable from 'xlsx/dist/cpexcel.full.mjs';
xlsx.set_cptable(cptable);
```

PR #2793, opened 2022-09-16
Add `types` to subpath exports in `package.json`

phaseOne:
When a project's `tsconfig.json` is configured to use ECMAScript modules and has `strict` or `noImplicitAny` set to `true`, TypeScript complains about being unable to find `xlsx`'s declaration file:
```
TSError: ⨯ Unable to compile TypeScript:
src/index.ts:1:23 - error TS7016: Could not find a declaration file for module 'xlsx'. '<OMITTED>/node_modules/xlsx/xlsx.mjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/xlsx` if it exists or add a new declaration (.d.ts) file containing `declare module 'xlsx';`

1 import * as XLSX from 'xlsx';
```

`tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "ES2015",
    "module": "Node16",
    "outDir": "./build",
    "sourceMap": true,
    "allowJs": true,
    "strict": true
  },
  "include": ["./src/**/*"]
}
```

The change in this PR should resolve the issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,22 +28,27 @@
 	"exports": {
 		".": {
 			"import": "./xlsx.mjs",
-			"require": "./xlsx.js"
+			"require": "./xlsx.js",
+			"types": "./types/index.d.ts"
 		},
 		"./xlsx.mjs": {
-			"import": "./xlsx.mjs"
+			"import": "./xlsx.mjs",
+			"types": "./types/index.d.ts"
 		},
 		"./dist/xlsx.zahl": {
 			"import": "./dist/xlsx.zahl.mjs",
-			"require": "./dist/xlsx.zahl.js"
+			"require": "./dist/xlsx.zahl.js",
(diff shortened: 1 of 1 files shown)


SheetJSDev:
`xlsx.zahl` exports a default string while the `cpexcel` scripts export the same shape as the main codepage library https://github.com/SheetJS/js-codepage/blob/master/types/index.d.ts .  For the purposes of this patch, just include the types for the root export and `xlsx.mjs`.

SheetJSDev:
[4dd092a](https://github.com/SheetJS/sheetjs/commit/4dd092a076e23c5ddc40ef1469af9d8578ed4e6a) cherry-picks the first two changes.

Is there a simple project template to verify correctness?

SheetJSDev:
Tested in a simple svelte+TS app and the types seem to work using the changes from 1491302.

PR #2807, opened 2022-10-24
Fix JSON.parse undefined error when o is function

mehrangulizadeh:
Prevent JSON.parse undefined error when 'o' variable is function


SheetJSDev:
Please raise an issue first with some way to reproduce the error.  Can you elicit this error with code that would pass a TS strict type check?

Issue #1105, opened 2018-05-14
Rounding does not always match Excel's display values


feverpitch-tom:
Hi there

For certain cell values the xlsx package yields different formatted display text than Excel.

A full example with a file is available at https://github.com/feverpitch-tom/xlsx-rounding-demo.

For manual reproduction:

Create a spreadsheet with the following three values:

- 1.225
- 100.665
- 656.685

These should be formatted to display at two decimal places, and consequently are displayed as follows in Excel:

- 1.23
- 100.67
- 652.69

When parsed by the xlsx package the w property for each cell, which represents the formatted value, is as follows:

- 1.23
- 100.67
- 652.68

The last value (C1 in the sheet) is 0.01 less than is displayed in Excel.

Thanks 

Tom

SheetJSDev:
Thanks for reporting!  Excel is storing the IEEE754 value 652.68499999999995, just shy of traditionally rounding to 652.69 😞

![597d1386-439d-4c05-b4d6-9fd2b201d00e](https://user-images.githubusercontent.com/6070939/40035579-5b600d5c-57d0-11e8-8782-754c465efc80.png)

feverpitch-tom:
I've submitted a PR with a potential fix - it'd be good to get your feedback @SheetJSDev

feverpitch-tom:
@SheetJSDev did you get a chance to look through that PR?  It'd be great to hear your views 😄

feverpitch-tom:
@SheetJSDev it's all going a bit over my head :-D 

Looking at those sheets and the `ssf` library i'm not sure there's anything we can do to resolve this :-(

yfengBTI:
Adding another case where rounding doesn't match up to Excel:

Excel rounds halves **away from zero**:
- 0.5 rounds to 1
- -0.5 rounds to -1

Whereas SheetJS rounds halves **up**:
- 0.5 rounds to 1
- -0.5 rounds to 0

PR #2809, opened 2022-11-20
fix-2824_numbers rounding

andrei-kav:
some kind of solution for this issue https://git.sheetjs.com/sheetjs/sheetjs/issues/2824


andrei-kav:
@SheetJSDev could you please take a look at it? Based on https://github.com/d3/d3-format/issues/27#issuecomment-262583417

SheetJSDev:
#1105 proposed a similar solution.  There's a longer answer which will be left in the tracking issue

PR #2811, opened 2022-11-29
fix: getTimeZoneOffset needs to return a double instead of integer at v8

queguangwei:
I found that the return time of the datenum method in Chrome browser is incorrect, consider customizing a getTimeZoneOffset function to fix Date.prototype.getTimezoneOffset() return error value

### DIFF
--- a/xlsx.js
+++ b/xlsx.js
@@ -3268,19 +3268,26 @@ var basedate = new Date(1899, 11, 30, 0, 0, 0); // 2209161600000
 function datenum(v, date1904) {
 	var epoch = v.getTime();
 	if(date1904) epoch -= 1462*24*60*60*1000;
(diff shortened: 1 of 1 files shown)


SheetJSDev:
This issue was reported to V8 years ago https://bugs.chromium.org/p/v8/issues/detail?id=7863 but it seems like there has been no movement (and unfortunately a large number of issues are reported here).

queguangwei:
I see, since v8 does not fix the problem, do we consider fixing the problem in the code.

koticmit:
@SheetJSDev, Would be nice if we get new release with latest commits. Waiting for once of the security fix ReDos in your package to be fixed. They say below that it is fixed in master branch but not yet released. Many thanks if you can do it.
image

PR #2813, opened 2023-01-01
fix(types/JSON2SheetOpts): extend OriginOption

chsdwn:


SheetJSDev:
The `origin` option was added to `sheet_add_json` (that is the old `SheetJSONOpts` interface).  The type was removed in https://git.sheetjs.com/sheetjs/sheetjs/commit/5141222c24f2ae457ff74a62d4b76efc9467b99b

PR #2814, opened 2023-01-18
Check for strings that may be interpreted as numbers (fixes #2609)

### LINKED ISSUE #2609: Auto converting string data with all integer to exponential.
While writing string data (example = "121212121212121" -> string of all integer characters) into excel.
while user download and see, the downloaded xl/CSV, the data is shown in exponential format. Is there a way to show the same data. without converting into  exponential in this case.

qburst-harikrishnanp:
fixes #2609


SheetJSDev:
Cell type checks are needed in both parts to determine whether the intention is to pass a number or a string.

[Please raise an issue in the main project repo](https://git.sheetjs.com/sheetjs/sheetjs/issues) with some examples to start the discussion

Issue #2752, opened 2022-08-02
sheet_to_csv skips first blank line with blankrows: true option


pibi:
**sheet_to_csv** skips the first blank line of a sheet, even with an explicit `{blankrows: true}` set.

## Code sample:

```
  var ws = XLSX.utils.aoa_to_sheet([
    [  ,   ,    ,    ,   ,   ,   ],
    ["S", "h", "e", "e", "t", "J", "S"],
    [  ,   ,    ,    ,   ,   ,   ],
    [  1,   2,    ,    ,   5,   6,   7],
    [  2,   3,    ,    ,   6,   7,   8],
    [  3,   4,    ,    ,   7,   8,   9],
    [  4,   5,   6,   7,   8,   9,   0]
  ]);
  
  console.log(XLSX.utils.sheet_to_csv(ws, {blankrows: true}));
```

## Result:
```
S,h,e,e,t,J,S
,,,,,,
1,2,,,5,6,7
2,3,,,6,7,8
3,4,,,7,8,9
4,5,6,7,8,9,0
```

## Expected:
```
,,,,,,
S,h,e,e,t,J,S
,,,,,,
1,2,,,5,6,7
2,3,,,6,7,8
3,4,,,7,8,9
4,5,6,7,8,9,0
```

SheetJSDev:
This is an `aoa_to_sheet` issue.  It will generate a sheet where the starting cell is A2 rather than A1.  See https://github.com/SheetJS/sheetjs/issues/2737 for more details on why the range is relevant.

The simplest fix is to pin the starting point to the origin in https://github.com/SheetJS/sheetjs/blob/master/bits/27_csfutils.js#L176 (we'll accept a PR):

```js
  if(range.s.c < 10000000) {
    if(range.s.c > _C) range.s.c = _C;
    if(range.s.r > _R) range.s.r = _R;
    ws['!ref'] = encode_range(range);
  }
```

anmol5varma:
Is the issue still open? I would like to take it up if that is the case.

anshul137:
Is the issue still open? I would like to take it up if that is the case.

PrathyushaModala:
Hi @pibi ,
I'd like to work on this, could you please assign me :)

naikmohit13:
is it still open?

bardia-rk:
i will work on it

sayantandasgupta:
[...]

PR #2815, opened 2023-02-25
Fix for Issue 2752 where blankrows:true skips first blank line

abhiram11:
This PR tries to solve #2752 which is also mentioned in PR #2776 but it did not pass all tests of a PR as set by SheetJS.

This PR consists of two commits:
1. Same code as mentioned in https://github.com/SheetJS/sheetjs/issues/2752#issuecomment-1203205189 
2. After running `yarn lint`, two other files were also updated, namely `xlsx.js` and `xlsx.flow.js`, which are also committed in this PR. These were missing in the code changes of PR #2776 .

Let's hope this time all tests cases are passed!

### DIFF
--- a/bits/27_csfutils.js
+++ b/bits/27_csfutils.js
@@ -173,7 +173,11 @@ function sheet_add_aoa(_ws/*:?Worksheet*/, data/*:AOA*/, opts/*:?any*/)/*:Worksh
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


abhiram11:
There seems to be plenty of test cases failing, please let me know if there is something specific that I'm doing wrong or is there some other reason itself!

PR #2824, opened 2023-08-04
remove property len on scbs method

JuanFRamirez:


JuanFRamirez:
remove property len on scbs method for testing purposes

